PR #11, opened 2021-01-18
First test with Nevergrad

teytaud:
This PR is experimental.
We replace the current hyperparameter sampling by LHS by an optimization by Nevergrad.
The code is not parallelel anymore: this is not for a deep reason, I just start learning R and did not know how to communicate between threads in the "foreach".

# Type of change
 * New feature: Possibility to use Nevergrad, making the whole process faster.

# How Has This Been Tested?
Checklist:
 * Fork the repo and create your branch from master.
 * If you've added code that should be tested, add tests.
 * If you've changed APIs, update the documentation.
 * Ensure the test suite passes.
 * Make sure your code lints.


source/fb_nextgen_mmm_v20.0.exec.R
```diff
@@ -82,7 +83,7 @@ set_modTrainSize <- 0.74 # 0.74 means taking 74% of data to train and 30% to tes
 
 ## set model core features
 adstock <- "geometric" # geometric or weibull . weibull is more flexible, yet has one more parameter and thus takes longer
-set_iter <- 50000 # We recommend to run at least 50k iteration at the beginning, when hyperparameter bounds are not optimised
+set_iter <- 100  #50000 # We recommend to run at least 50k iteration at the beginning, when hyperparameter bounds are not optimised
```
teytaud, line 86: temporary, for playing quickly with the code.

source/fb_nextgen_mmm_v20.0.func.R
```diff
@@ -925,6 +946,7 @@ f.mmm <- function(...
       )
       
       setTxtProgressBar(pb, i)
+      optimizer$tell(nevergrad_hp, mape)
```
teytaud, line 949: we minimize "mape"

teytaud:
An enormous "thank you" to Gufeng, Jeremy, and the other Jeremy :-)

I compared Nevergrad, one core, 1000 trials, to the original LHS sampling with 50 000 trials.
This corresponds to 50x less budget.
This is the first experiment, so maybe I miss a point; feedbacks are very welcome.

Pros:
- Nevergrad gets better results with 1 000 trials than LHS with 50 000 trials. I did not try to find how much LHS trials we need for competing with 1000 Nevergrad trials, but 1000 nevergrad is better by a vast margin so I expect a speed-up closer to 300 than to 50 for the default problem size.
- The change is quite light in terms of number of lines.
- The best performing methods from Nevergrad are those we could have conjectured from the description of the problem.

Cons:
- I failed to run the multithreaded mode, as my code was not directly compliant with the multithreading in R. This should be fixed but this is my first code in R since my childhood so for this first draft I did not do the parallelization. Please note that given the speed-up (significantly more than 50, as we get clearly more), the change is still beneficial for at least 50 cores and probably much more.
- This adds a "reticulate" dependency. Reticulate is however a quite stable package.
- I did not lint properly, I am a R beginner, but I'm happy to do so as soon as I learn a bit more :-)


single core, 1000 trials, 1 epoch, performance of various methods ("mape" criterion, the lower the better):

lhs: 11.4552                                                               <===== test original code with budget 1000, i.e. equivalent
[1] "SQP" [1] 1000 [1] "->" [1] 11.31854
lhs: 11.28726                                                              <===== test original code with budget 1000, i.e. equivalent
[1] "ScrHammersleySearch" [1] 1000 [1] "->" [1] 11.22211
[1] "cGA" [1] 1000 [1] "->" [1] 11.13627
[1] "OnePlusOne" [1] 1000 [1] "->" [1] 11.11502
[1] "Powell" [1] 1000 [1] "->" [1] 11.0734
[1] "ScrHaltonSearch" [1] 1000 [1] "->" [1] 11.00924
lhs: 10.97486                                                              <===== test original code with budget 1000, i.e. equivalent
[1] "CMA" [1] 1000 [1] "->" [1] 10.96881
[1] "RandomSearch" [1] 1000 [1] "->" [1] 10.94271
lhs10000: 10.88336`                                                              <===== this one is run with budget 10 000 (original code)
lhs50000: 10.871                                                                    <===== this one is run with budget 50 000 (original code)
[1] "TwoPointsDE" [1] 1000 [1] "->" [1] 10.77919
[1] "DE" [1] 1000 [1] "->" [1] 10.62383
[1] "Cobyla" [1] 1000 [1] "->" [1] 10.61805
[1] "DiscreteOnePlusOne" [1] 1000 [1] "->" [1] 10.57731
[1] "NaiveTBPSA" [1] 1000 [1] "->" [1] 10.56018
[1] "DoubleFastGADiscreteOnePlusOne" [1] 1000 [1] "->" [1] 10.53541
[1] "DoubleFastGADiscreteOnePlusOne" [1] 1000 [1] "->" [1] 10.2304
[1] "DiscreteOnePlusOne" [1] 1000 [1] "->" [1] 10.04268
MetaModel --> crashes ?

Feedback very welcome !!

source/fb_nextgen_mmm_v20.0.func.R
```diff
@@ -741,16 +747,25 @@ f.mmm <- function(...
 
   ## set paralle backend
   cat("\nRunning", iterRS,"random search trails with",lambda.n,"trails lambda cross-validation each on",set_cores,"cores...\n")
-  cl <- makeCluster(set_cores)# makeSOCKcluster(set_cores) #makeCluster(set_cores)
-  registerDoSNOW(cl)
+  #cl <- makeCluster(set_cores)# makeSOCKcluster(set_cores) #makeCluster(set_cores)
```
Leonelsentana, line 750: This is the part we need to solve for parallel computing on nevergrad

source/fb_nextgen_mmm_v20.0.func.R
```diff
@@ -741,16 +747,25 @@ f.mmm <- function(...
...
   t0 <- Sys.time()
+
+  # Creating an optimizer.
+  my_tuple <- tuple(length(hyperParams[[1]]))
+  instrumentation <- ng$p$Array(shape=my_tuple)
+  instrumentation$set_bounds(0., 1.)
+  optimizer <-  ng$optimizers$registry["DiscreteOnePlusOne"](instrumentation)  # length(hyperParams[[1]]))
+  # Creating an hyperparameter vector to be used in the next learning.
+
   sysTimeDopar <- system.time({
```
Leonelsentana, line 786: It seems he is computing in parallel anyways?

source/fb_nextgen_mmm_v20.0.func.R
```diff
@@ -774,9 +789,15 @@ f.mmm <- function(...
       #####################################
       #### Get hyperparameter sample
       
+      # Let us create a vector of hyperparameters using Nevergrad.
+      nevergrad_hp <- optimizer$ask()
+      # Now we must cast it to the bounds.
+      # ... TODO ! This is missing
+
       hypParamSam <-  sapply(hyperParams, function(x) {if (length(x) > 1) { x[i] } else {x} }); hypParamSam
```
gufengzhou, line 811: hypParamSam needs to be generated from your 1+1 vector nevergrad_hp instead of hyperParams from my lhs. Also the next row: hypParamSamName needs to have the same hypPar names as your nevergrad_hp. Then we're good to go:)

teytaud: At line 853 below we do fill hypParamSam with nevergrad_hp, right ? So I don't get it.
I'm new to the R syntax so I might be wrong :-)

source/fb_nextgen_mmm_v20.0.func.R
teytaud: Yes!

source/fb_nextgen_mmm_v20.0.func.R
teytaud: I think it does not.
It does not mean that it's necessarily an issue - the speed-up due to switching to Nevergrad is much greater than the parallel speed-up. However, as Nevergrad is fundamentally able to run in parallel (including asynchronously) it would be bettter to make this parallel. I would find that easy in Python but I need to learn a bit of R syntax for doing it.

What we need to do (for 10 cores) if we don't care about asynchronous optimization:
- add an argument "num_workers = 10" in line 753 "optimizer <- ...., budget=iterRS", so that the algorithm knows it must be parallel (some algorithms don't care)
- do 10 times "ask" (sequentially, this is anyway super fast)
- compute in parallel the 10 "mape" corresponding to those 10 nevergrad_hp
- do 10 times "tell" (sequentially, this is anyway super fast)

source/fb_nextgen_mmm_v20.0.func.R
gufengzhou: line 853? I can't see it somehow :)

teytaud: From the oral conversation today I guess the code is actually ok.
Next versions:
- including two-objective optimization
- including constraints
- parallelizing, maybe

source/fb_nextgen_mmm_v20.0.func.R
Leonelsentana: Great Oli thanks so much for your explanation. One more for the backlog! Maybe we can add this once we tested the objective function with more than one variable.

PR #106, opened 2021-07-05
Improved f.robyn.func.R (R session aborted fixed bug )

AnisLEBIK:
# Contributing to FB NextGen MMM R script

Before this improvement, each time restarting a modeling session a second time, the R session crashed because the parallelization was not correctly closed.

These changes therefore give :

    - The ability to launch several modelizations in a row without crashing R session.
    - The possibility of looping on several models at once and correctly close our parallelizations after modeling

### DIFF
--- a/source/fb_robyn.func.R
+++ b/source/fb_robyn.func.R
@@ -1021,7 +1021,10 @@ f.mmm <- function(...
       decomp.rssd.collect <- c()
       best_mape <- Inf
       closeAllConnections()
-      registerDoParallel(set_cores)  #registerDoParallel(cores=set_cores)
+      al <-makeCluster(set_cores) # Create the clusters
+      registerDoParallel(al) # Register the parallel backend 
+      # registerDoParallel(set_cores)  
+      #registerDoParallel(cores=set_cores)
       getDoParWorkers()
       doparCollect <- foreach (
         i = 1:iterPar
@@ -1258,7 +1261,8 @@ f.mmm <- function(...
         return(resultCollect)
       } # end dopar
       ## end parallel
-      
+      stopCluster(al)#
+                                                                    
       nrmse.coolect <- sapply(doparCollect, function(x) x$nrmse)
       decomp.rssd.coolect <- sapply(doparCollect, function(x) x$decomp.rssd)
       mape.lift.coolect <- sapply(doparCollect, function(x) x$mape.lift)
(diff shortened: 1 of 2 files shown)


rubenharutyunyan:
Thank you Anis for this modification, I needed to model several models in a row and the Session R "crashe" was quite disturbing. I searched for a long time but I did not find the solution.

gufengzhou:
Hi all, thanks for this PR! We're hearing about this bug for the first time and will examine the changes. Rubenharutyunyan can you confirm that this fixes your problem completely?

rubenharutyunyan:
> Hi all, thanks for this PR! We're hearing about this bug for the first time and will examine the changes. Rubenharutyunyan can you confirm that this fixes your problem completely?

Hello, for personal needs, I created loops to model multiple regions at once but since adding nevergrad my sessions consistently crash after the first model. I also found that running the same model multiple times also caused the R session crash after the first model. With this modification, I had no problem for the 2 cases mentioned above.

gufengzhou:
Thank you for the feedback! We're freezing the code this and next week due to internal process and will examine and implement this PR asap.

gufengzhou:
Hi, we've noticed that using makeCluster() and stopCluster() slows down the parallel performance on my side quite dramatically. According to this [https://stackoverflow.com/questions/28829300/doparallel-cluster-vs-cores](url) thread, makeCluster/stopCluster uses the SNOW backend only that is intended for Windows and not Unix-like systems, while registerDoParallel(cores) chooses the right backend depending on OS. I'd reverse this commit until we find a more universal solution. Would be great if you can try updating your R to 4.1.0 and try using registerDoParallel() again to see if that works better.

AnisLEBIK:
Hi @gufengzhou  i  updated R to 4.1.0 and used registerDoParallel() : 

R version

I have the same issue, R session crashes when i launch a second model in a row:

R session aborted 2

gufengzhou:
Are you on windows or mac/other OS? Local/cloud?
I'm seeing in your screenshot that it crashes after printing "plotting 22 pareto....", which indicates that the parallelisation is actually already done.

AnisLEBIK:
I'm working on Windows and on Local. Yes it crashes after parallelisation. But the problem is caused by a bad closure at the end of the parallelization. And at the time of plotting the results, the memory saturates due to a bad closing of the connections : 

R session restart

So I think registerDoParallel () does not allow full and correct shutdown of parallelization on Windows. Indeed, some parallel computing going on in the background that is not getting cleaned up fully between runs. Unlike make / close Cluster which allows a better control of parallelization.

gufengzhou:
According to this documentation https://www.rdocumentation.org/packages/doParallel/versions/1.0.16/topics/registerDoParallel, the function stopImplicitCluster can close the connections implicitly. Can you try it out and let me know?

gufengzhou:
Also interesting for you, in this issue https://github.com/facebookexperimental/Robyn/issues/118, the user sees increase of performance when switching back from makeCluster. He probably works with Unix-like systems. My goal is to have a solution that works for both OS

AnisLEBIK:
Hi @gufengzhou , I have tried implementing the stopImplicitcluster in the code. It works but it slows down the processing time on Windows. Furthermore, I managed to find a solution that suits both OS and increases performance while decreasing processing time for both systems.
Could you process the last pull request #122 I posted so that I can suggest this improvement ?

Issue #108, opened 2021-07-07
Robyn does not work on Databricks notebooks


danzafar:
# Contributing to FB NextGen MMM R script

# Issue

Robyn does not work on Databricks notebooks. Due to this line:

https://github.com/facebookexperimental/Robyn/blob/master/source/fb_robyn.func.R#L1023

`closeAllConnections()` will close the socket that sends data to/from R REPL stdout/stderr and the Notebook.

I mitigated the issue by simply omitting the line before sourcing that script:
```
%sh
sed -i /closeAllConnections/d ~/robyn/source/fb_robyn.func.R
```

Here are the connections that exist by default in the notebooks:
```
> showConnections()
  description         class      mode text     isopen   can read can write
3 "->localhost:36985" "sockconn" "a+" "text"   "opened" "yes"    "yes"    
4 "->localhost:36263" "sockconn" "a+" "text"   "opened" "yes"    "yes"    
5 "->localhost:35161" "sockconn" "wb" "binary" "opened" "yes"    "yes"   
```
As long as these are not deleted, it will work.

# Steps to Reproduce

Running any minimal version of `f.robyn()` fails. You can replicate the issue by just running the problematic line:

`closeAllConnections()`

# Expected Behavior

It would be nice for Databricks users to be able to experiment with Robyn.

# Actual Results

After running the problematic code, any subsequent notebook operations yeild:
```
[Fetch Stdout Timeout]
Error in writeLines(DB_STREAM_BEACON_STRING_END__, DATABRICKS_STDOUTCONNSTREAM__) : 
  invalid connection
Some(<code style = 'font-size:10p'>  </code>)
```

# Environment

R version: 4.0.4, DBR 8.3 ML.

gufengzhou:
Hi, thanks for raising this. To be honest I have no experience with databricks. We'll see how to avoid closing all connections and change it in the next update.

le1tz3y:
I am facing the same issue as described by @danzafa. I tried removing `closeAllConnections ()` before sourcing, but I am still seeing the error.

danzafar:
@gufengzhou - can you link the PR that fixed this? Otherwise, it is disingenuous to close out the issue.

gufengzhou:
Hey guys, my mac is crashing with your fix with makeCluster and closeCluster. I'm asking others to try it out too. Would you  please recheck your fix for non-databricks environment? Also have you tried doParallel::stopImplicitCluster() ?

[...]

PR #126, opened 2021-08-01
Improvement of Robyn code

AnisLEBIK:
# Contributing to FB NextGen MMM R script

## Fixes 
- Resolve the problem of R session crashing  after having launched two models in a row on WIndows.
- Drastically reduces model processing time on both operating systems.
- Could potentially fix the problem encountered on Databricks notebooks : #108 

##  Type of change

- Replace doParallel by registerdoFuture and dopar by dorng
- Remove CloseAllConnections()

# How Has This Been Tested?

PC Test Environement 1 : Desktop computer 16 cores / Windows system

+ RegisterDoParallel() :
doParallel_16cores

+ RegisterDoFuture() :
	
doFuture_16 cores

---------------------------------------------------------------------------------------------------------------------------------------------

PC Test Environement 2 : Laptop 8 cores / Windows system	

+ RegisterDoParallel() :
DoregisterParallel_8cores

+ RegisterDoFuture(): 
DoFuture_8cores
	
---------------------------------------------------------------------------------------------------------------------------------------------

PC Test Environement 3 : Laptop 4 cores / MAC OS system

+ RegisterDoParallel() :
	
doParallel_MAC OS_ 4 cores

+ RegisterDoFuture() :
doFuture_MAC OS_ 4 cores

## SUMMARY 

Time per trial
                          
Total time

### DIFF
--- a/source/fb_robyn.func.R
+++ b/source/fb_robyn.func.R
@@ -1020,8 +1020,8 @@ f.mmm <- function(...
       nrmse.collect <- c()
       decomp.rssd.collect <- c()
       best_mape <- Inf
-      closeAllConnections()
-      registerDoParallel(set_cores)  #registerDoParallel(cores=set_cores)
+      registerDoFuture()
+      plan(multicore)
       getDoParWorkers()
       doparCollect <- foreach (
         i = 1:iterPar
@@ -1038,7 +1038,7 @@ f.mmm <- function(...
                         ,'data.table'
         )
         #, .options.snow = opts
-      )  %dopar%  {
+      )  %dorng%  {
         
         t1 <- Sys.time()
         


gufengzhou:
thanks for committing this! Looks very promising esp on windows. I just tested it on my 8-core Macbook Pro (Big Sur) and doFuture it's very close or slightly slower than doParallel somehow. I'm wondering why you see such improvement on your mac and I don't. 

Also, when reading ?multicore as in plan(multicore), it says "Not all operating systems support process forking and thereby not multicore futures. For instance, forking is not supported on Microsoft Windows....Trying to create multicore futures on non-supported systems or when forking is disabled will result in multicore futures falling back to becoming sequential futures." Therefore I don't quite understand why you see such improvement on Windows using multicore. Do you have an idea?

AnisLEBIK:
Hi , I couldn't explain why it doesn't reduce processing time on your computer. Before comitting this PR i launched several times to make sure it worked perfectly and reduced the model's processing time.

Indeed, plan(multicore) doesn't work on Windows, I left it for MAC OS environments. The code does not take this into account and implicitly uses plan (sequential).

gufengzhou:
hey, thanks for digging into this and i'll reply here just so others can see it. As comparison, I've tested these combinations for you as reference. Quite strange that I can't see the improvement like yours.
image
Regarding all OS, it's actually easy to use the output from .Platform$OS.type ("windows" for windows and "unix" for the rest) to create a OS-specific function to use multisession for windows and multicore for unix, if I understood [this](https://github.com/HenrikBengtsson/doFuture) correctly?

AnisLEBIK:
Hey, by analyzing the results of your comparisons i could see that there is no difference when you use 6 cores or 1 core with registerDoParallel which seems strange to me. I retested to compare both parallelization methods . 

results


 I tried using more cores than my pc's capacity and it worked. Could you test it? 

In terms of specifying the type of plan I had thought about it at the beginning, but registerDoFuture () with plan (session) does the same processing as registerDoParallel () and therefore takes a long time processing. The code should work fine as it is now, or in a pinch we could specify  plan (sequential) for Windows and plan(multicore) for Unix. But it remains optional.

AnisLEBIK:
Hey, by analyzing the results of your comparisons i could see that there is no difference when you use 6 cores or 1 core with registerDoParallel which seems strange to me. I retested to compare both parallelization methods . 

results



 I tried using more cores than my pc's capacity and it worked. Could you test it? 

In terms of specifying the type of plan I had thought about it at the beginning, but registerDoFuture () with plan (session) does the same processing as registerDoParallel () and therefore takes a long time processing. The code should work fine as it is now, or in a pinch we could specify  plan (sequential) for Windows and plan(multicore) for Unix. But it remains optional.

gufengzhou:
how do you specify cores when using registerDoFuture() and plan(multicore)?  like this plan(multicore(workers = 6)) or how do you do it?

AnisLEBIK:
In set_cores

AnisLEBIK:
Actually specifying set_cores will be more for nevergrad optimizer: 
optimizer <-  ng$optimizers$registry[optimizer_name](instrumentation, budget=iterTotal, num_workers=set_cores)

gufengzhou:
yes, when you only use plan(multicore) instead of plan(multicore(workers=set_cores)), multicore will take default value which is availableCores(). It's better to align it with Nevergrad. Thus I recommend you change it to plan(multicore(workers=set_cores)). 
image

Strangely, I still don't see improvement on my laptop, while my colleagues do see improvement. 
image

In general, without spending too much time on it, I do think that your change will improve experience for most users. Can you combine the two PRs into one, also specifying workers in multicore? then we can merge it.

AnisLEBIK:
Changing to plan(multicore(workers=set_cores)) causes an error on windows : Error in UseMethod("tweak") :
  no applicable method for 'tweak' applied to an object of class "c('SequentialFuture', 'UniprocessFuture', 'Future', 'environment')". That's why I made an OS-specific function to handle both systems.

gufengzhou:
An update on this: we have an user that is "using R jupyter notebook within JupyterLab on AI Platform with 64 cores." and is saying that the doFuture update has increased her runtime by like 5x compared to doParallel...any idea?

AnisLEBIK:
Hi, I have never used such a setup so I have no idea why the processing time is slower.

PR #136, opened 2021-08-17
Update step-by-step-guide.mdx

JavOrraca:
This is a basic documentation update to the Step-by-step guide to highlight two sets of inline code syntax under Set Global Parameters (item # 6)


gufengzhou:
Thanks for the improvement. We're updating documentation to match the next release in general and all current content will be reworked. So there's no need to fix this at the moment.

PR #217, opened 2021-11-12
Multi-core plotting Pareto optimum models

andrey-legayev:
# Project Robyn

I've found that Pareto plotting takes significant amount of time (we use pareto_fronts=3, which results in having 80 models).
I've refactored the code to run plotting in parallel, while keeping progress bar in R Console.
The solution works in multi-core mode on Windows 10.

Results of benchmarking: new version is 57 sec faster on 20 models (running on 6 cores)

# Type of change

- Performance optimization (non-breaking change)

# How Has This Been Tested?

Tested by running Robyn MMM on our internal data: I've compared outputs of unmodified version of Robyn vs. my changes.
Environment: R version 4.1.1 / Python 3.9.7 / Windows 10

Result of testing: there are no changes in data and in PNG plots.
Total execution time is ~1 min smaller.


gufengzhou:
Good idea actually. Although we've moved away from doSNOW to doFUTURE because of compatibility for difference OS.

andrey-legayev:
Hi @gufengzhou,

I used doSNOW only to have ability to update progress bar, no other backend allows such functionality.
So, if you would like to speed up rendering all models by using many cores and show progress bar in console, then SNOW is the only solution.

BTW What is the list of officially supported platforms?

**Other backends**
From the source code I understand that you moved from doSnow/doParallel to doFuture
Note: doFuture is actually used now: [R/R/model.R](https://github.com/facebookexperimental/Robyn/blob/fca1d04dec69dc2be75941dc354482a223607438/R/R/model.R)

It looks to me as you tried to use doParallel, but then commented it out because of some issues.
I don't see any history of that change in Git, the whole file has been added in one commit: 66015701630b8c1aa7589fb54ae296c4cc919f73

I read about doParallel backend and tested it on Windows and Linux, this is what I've found so far:
1. On Unix-type OS it works very similar to doFuture (tested on Ubuntu)
In this mode it creates additional threads and run parallel code in threads. Threads share the same variables, no data copying is needed.
```
> .Platform$OS.type
[1] "unix"
> registerDoParallel(4)
> print(paste(getDoParName(), getDoParVersion(), getDoParWorkers()))
[1] "doParallelMC 1.0.16 4"
```
2. But on Windows it automatically selects SNOW backend:
```
> .Platform$OS.type
[1] "windows"
> registerDoParallel(4)
> print(paste(getDoParName(), getDoParVersion(), getDoParWorkers()))
[1] "doParallelSNOW 1.0.16 4"
```
SNOW backend creates processes instead of threads which means much higher overhead in general.
In case of Robyn MMM parallel code operates with big arrays / data.frames from the function robyn_run()
In order to run it in multi-process mode SNOW backend needs to copy those variables to every process which it creates.
It happens 2000 times. I won't be surprised if parallel mode is slower than sequential.
The code requires significant refactoring in order to make it working fast enough in SNOW mode.
I would suggest to run parallel section not per iteration, but per 50 or 100 iterations. It would reduce overhead for Unix-systems as well.

Bottom line:
 - doParallel works fine on Unix systems - it should show the same performance as doFuture
     - for Windows registerDoSEQ() can be used to make it work as it works with plan(sequential)
     - I tested it, it works fine in Linux and Win10: 4bc7f3cfbd5c554b3210108fe42bade0ee8c8125
 - doSnow (or doParallel on windows) have very big specifics, but it should work just fine for PNG rendering
     - I tried to run Robyn with registerDoParallel() on windows and it resulted in huge memory leak and many not killed processes
 - showing progress bar is nice, but tricky for parallel computations - it has own coplexities, for now it works fine with SNOW

andrey-legayev:
BTW The change which I did - moving setup of muti-core mode to be above for loop gives feasible improvement.
I'll submit pull request for that small change.

My test runs on 200 iterations and 4 cores shown 9.5 sec improvement (0.71 -> 0.55 min)
I tested it in Ubuntu / VirtualBox under Win10 host.

It's also visible that new version spends less time in single-threaded mode (setup of foreach-backend) which results in higher CPU usage overall:
image

gufengzhou:
hey, I've tested your change from doFuture to doParallel in the other PR https://github.com/facebookexperimental/Robyn/commit/4bc7f3cfbd5c554b3210108fe42bade0ee8c8125 as well as this PR with parallel plotting. It works great on my MacBook and it's visibly faster, both in modelling and plotting iteration. Great solution, thanks!

One more question: looking at both PRs, we'll be having registerDoParallel for the main modelling loop and registerDoSNOW for the plotting, because for the latter one there's no other option for progress bar? I think it's good to have the same approach for both parallel loops. I tend to agree to using registerDoParallel for both, because we'll be having one less dependency. Also the progress bar in the plotting in "batched", too, same as in the modelling loop, meaning we could use the same progress bar approach in both places. What do you think? 

Also it probably makes sense to combine these two PRs if you're reworking them anyways?

andrey-legayev:
Sure, using registerDoParallel everywhere makes sense. I used SNOW only because it has a nice ability to update progress bar.
By the way registerDoParallel will work even faster than SNOW on Mac.

So, I'll do following changes later on this week:
 - switch pareto plots to registerDoParallel with updating progress bar after each batch
 - merge two pull requests into one

andrey-legayev:
Done - pls review.
Tested on Windows and Linux.

The time diff on Linux is significant (4 cores):
 - main modelling loop on just 100 iterations: 0.33 -> 0.25 min
 - pareto plotting for 24 models: 3.3 -> 1.52 min

On windows time saving mainly happens only because of Pareto parallelization.

gufengzhou:
Hey, I've just tested it. The main model loops goes faster now: for 100 iters  0.25 -> 0.16min on my Mac, great improvement!
However it stucks at plotting and never moves forward as in screenshot. 

In your suggestion we have 3 foreach loop now in Robyn. I've figured that if I remove the stopImplicitCluster after the resp_collect parallel loop, and also remove the registerDoParallel(InputCollect$cores) before the plotting parallel loop, then it works for me. It seems like it's having problem on my machine when stopping it twice in a function? Although this would be strange, because stopImplicitCluster is used after the modelling loop too. What do you think? Can you test it on your side?

image

andrey-legayev:
Very strange... I will try to reproduce it.
How do you run it: RStudio / R console  / etc ?

gufengzhou:
thanks. you said you've tested it on windows & linux but not macOS, right? I'm on RStudio on macOS Big Sur 11.6

andrey-legayev:
Thanks for details.

I've tested it under Windows 11 and Ubuntu 20.04. 
I perform tests by preparing R script to run everything on top of specific Robyn branch. And then run it from R Console on different platforms.
RStudio has own specifics - what's why it's important how code is executed.

I've just tried to run it from RStudio / Ubuntu - it works fine.
I'll try it on Mac in office later this week.

Do you use Robyn/demo/debug.R in some way for test runs? I didn't figure out how to use it so far

andrey-legayev:
Good news - it's reproducible on our old MacMini

**Results of debug:**
ggsave() gets stuck on MacOS in case of process forking. It's getting into deadlock because graphical device descriptor is getting cloned together with other process memory.
It happens only on mac and it's not reproducible on Linux. 
The reason for this: saving plots into PNG or other format is implemented by using proprietary graphical library of corresponding OS. I.e. any R plotting may have own low level specifics on different platforms.

This bug is not related to RStudio. It's reproducible in R Console on MacOS.

BTW, interesting fact: I've found that PNG files from the same code on Windows and Linux has different size in KB.
Probably this is the result of using different PNG libraries to render plots into file.

**Solution:**
The only suitable solution is to disable multi-core rendering for MacOS - pls review my changes.
Tested on Win11, Ubuntu 20.04, MacOS Big Sur 11.6

gufengzhou:
Thanks for the very useful insight! I didn't know about that. 

We're almost there:) In your latest commit, you create the parallel condition as ` parallel_plotting <- Sys.info()["sysname"] != "Darwin"` for plotting, meaning Windows & Linux will use `registerDoParallel(InputCollect$cores)`, while MacOS will use `registerDoSEQ()`. This makes sense for MacOS based on your finding above, but Windows will also use `registerDoParallel`. Compared to this, when you create `parallel_processing <- .Platform$OS.type == "unix"` for the other 2 parallel places, MacOS and Linux will use `registerDoParallel(InputCollect$cores)`, while Windows uses `registerDoSEQ()`, and you also print that "windows fall back to 1 core". 

Is this true though? According to [this](https://stackoverflow.com/questions/45819337/option-cores-from-package-doparallel-useless-on-windows/45823182) and [this](http://users.iems.northwestern.edu/~nelsonb/Masterclass/gettingstartedParallel.pdf), snow-like functionality is used on windows (while multicore-like functionality on unix). If I understood it right, we should use `cl <- makeCluster(InputCollect$cores)` and `registerDoParallel(cl)` for windows and `registerDoParallel(InputCollect$cores)` for unix for `parallel_processing`? For `parallel_plotting`, the extra condition for MacOS has to be added then. Does it make sense?

andrey-legayev:
> We're almost there:)

Ok :-) See my answers below - 

**Parallel/SEQ mode**
Conditions ```!Win``` and ```!MacOS (Darwin)``` are done on purpose:
 - main modelling loop:
     - Mac/Linux - Multi-core mode brings performance benefits
     - Windows - using SNOW cluster brings so much overhead that it works even slower than sequential mode, it happens because parallel section uses huge data tables from parent frame
 - pareto plotting:
     - Mac - cannot do in parallel mode, because of ggplot issue with cloned graphical dev descriptors
     - Linux - seems to work fine
     - Windows - works fine, using SNOW is faster than sequential code

> and you also print that "windows fall back to 1 core".

I added this change to avoid confusion that you run it on 6-8 cores, but waiting for 30 min, while it's actually in sequential mode.


**MacOS detection**
I checked data output of .Platform$OS.type and Sys.info() and decided to use ```Sys.info()["sysname"] != "Darwin"``` for !Mac condition. 
Honestly - I'm not sure this is the best condition to detect MacOS in R.

**SNOW Clusters**
```registerDoParallel(InputCollect$cores)``` works fine on all platforms with following difference:
 - on Unix-type it uses doParallelMC backend
 - on Windows it uses doParallelSnow

There is no need to create cluster manually with cores=N - it automatically forwards amount for cores to Snow cluster which is created implicitly - see below
```
### Windows 
> registerDoParallel(4)
> getDoParName()
[1] "doParallelSNOW"
> getDoParWorkers()
[1] 4

### Linux (Mac should show the same)
> registerDoParallel(4)
> getDoParName()
[1] "doParallelMC"
> getDoParWorkers()
[1] 4
```
I also checked it in [source code of doParallel](https://rdrr.io/cran/doParallel/src/R/doParallel.R)
image

Also I can ensure you that my CPU on Win-machine is very loaded when it renders Pareto plots :)
It's spiky, because it works in batches:
image

**Older versions of R**
I must say that I tested doParallel functionality and reviewed source code of the latest package version (R v4.1.1, v4.1.2).
It is possible that it had different behavior in previous versions.

andrey-legayev:
that was fast! thanks :)

gufengzhou:
You've already provided the reason for registerDoSEQ instead of registerDoParallel for windows in main loop before. Sorry I oversaw that and thanks for repeating. Thank you very much for this PR! Well done!

gufengzhou:
FYI version is updated to v3.4.8 with some minor changes + fix typo 327f7328f47a23912ce4082ef67cbc1b7d508905

PR #220, opened 2021-11-18
Fix error message when no organic vars were provided

andrey-legayev:
Our team found following error message if we run RobynMMM without organic vars (empty vector in parameters)

Error message:
```
>>> Plotting 17 Pareto optimum models...
|====
Error in `[.data.table`(dt_transformSpend, , `:=`((InputCollect$organic_vars),  (run_robyn_mmm.R#209): LHS of := isn't column names ('character') or positions ('integer' or 'numeric')
```

Fix is quite simple - please review it.

The change has been tested on real production data.

### DIFF
--- a/R/R/inputs.R
+++ b/R/R/inputs.R
@@ -565,7 +565,7 @@ robyn_engineering <- function(InputCollect, ...) {
   ################################################################
   #### Obtain prophet trend, seasonality and change-points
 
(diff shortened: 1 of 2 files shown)


gufengzhou:
Hey, thanks for raising this too. It might makes sense to also fix the same condition for prophet_vars, context_vars & factor_vars. Can you also implement that and re-submit?

PR #222, opened 2021-11-19
robyn_run() performance optimization by switch to doParallel

andrey-legayev:
This change improves multi-core performance of robyn_run() by switching from doFuture to doParallel foreach-backend.
I also moved doParallel cluster initialization to be before for(iterations) loop to reduce overhead.

# Type of change

- Performance Optimization

# How Has This Been Tested?
## Performance tests

Results of benchmarking of robyn_run() with iterations=1000 on real-life dataset - 
(historical data for 1000 days, modelling period: 30 days, 13 configured channels)
 - MacOS - 22% faster (11.57 -> 8.95min)
 - Linux - 24% faster (7.91 -> 5.94min)
 - Windows - 9% faster (5.37 -> 4.89min)
Note: Windows runs code on 1 core, but there is improvement in overhead introduced by multicore setup

Test Environment Specs:
 - MacOS - R 4.1.1, Python 3.9.9 (very old MacMini, 2-core Core i5, 8GB RAM)
 - Linux - R 4.1.2, Python 3.9.8, XUbuntu 20.04 (virtual machine under VirtualBox, 4 CPU cores, 12GB RAM)
 - Windows - R 4.1.1, python 3.9.8, Windows 11

# RStudio compatibility
I performed test run of the code in RStudio/Mac to make sure it works fine and doesn't cause issues in RStudio. No issues were observed. 
I also tried to google for RStudio + doParallel issues, similar to doFuture - so far, I didn't find any popular bug reports.


gufengzhou:
Thanks for this. Actually we've changed from doparallel to dofuture a while ago due to bad performance on other platforms. Have you tested it on different OSs?

andrey-legayev:
I've found old issue #108  - just added there long explanation why it won't fail :)

I did two parts of testing:
 - benchmarking on Mac/Linux/Windows
 - checking it in RStudio on Mac - I didn't find any issues

gufengzhou:
Sounds great! I'll spend time to test it next week. Please be patient that we try to be as thorough as possible to not break someone's pipeline

andrey-legayev:
BTW do you have a document with list of supported platforms and R versions?

gufengzhou:
Regarding version, it's specified in DESCRIPTION that R>=3.5. Platform/OS support is something we need to improve. We know that there're all kinds of users from Mac over Windows to various platforms (Databricks, Sagemaker etc.) who run Robyn successfully. The major roadblock for a seamless deployment is reticulate/Nevergrad installation. The Parallelism used to cause some problems too. But since we switched to doFuture it's been stable, regardless of performance

andrey-legayev:
These changes were finalized and merged in #217

PR #356, opened 2022-03-23
Fix log message "Using custom prophet parameters"

andrey-legayev:
# Project Robyn

When I submit custom Prophet parameters to robyn_input() as shown below:
```R
InputCollect <- robyn_inputs(
  ...
  , seasonality.mode = "multiplicative"
  ...
)
```

**Actual result** 
I see following log message without any custom parameters: 
`Using custom prophet parameters:`

**Expected result**  
`Using custom prophet parameters: seasonality.mode`

Pls review this tiny fix for the log message.

# Type of change

- Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

I run fixed version of Robyn as part of our internal project - everything works fine, log message looks as it should be.
No other issues observed.

### DIFF
--- a/R/R/inputs.R
+++ b/R/R/inputs.R
@@ -628,7 +628,7 @@ robyn_engineering <- function(x, ...) {
       c("", "..."))
     prophet_custom_args <- setdiff(names(custom_params), robyn_args)
     if (length(prophet_custom_args)>0)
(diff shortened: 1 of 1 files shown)


laresbernardo:
Good catch. Thanks for the fix @andrey-legayev

PR #399, opened 2022-05-31
refactor: modify inflexion computation for precision and simplicity

mast4461:
# Project Robyn

I tried to recreate the computation of the Hill saturation function, and found that Robyn's code for it seemed unnecessarily convoluted and wasteful.

Why is `range(x)` computed twice? (Maybe the R compiler is smart enough to just do it once anyway?)
Why create a linear sequence (and why with 100 steps in particular?) when it is immediately reduced to a scalar with `quantile`?
Why round the result at all, and why round to four decimals in particular?

I think all these things are unnecessary, and rewrote the computation to use a simple linear interpolation instead. I also renamed the variable to reflect it's content.

# Type of change

- Code cleanup

# How Has This Been Tested?

By logical reasoning and by manual comparison of the original and proposed computations on the same test data.

### DIFF
--- a/R/R/transformation.R
+++ b/R/R/transformation.R
@@ -169,11 +169,11 @@ adstock_weibull <- function(x, shape, scale, windlen = length(x), type = "CDF") 
 #' @return Numeric values. Transformed values.
 #' @export
 saturation_hill <- function(x, alpha, gamma, x_marginal = NULL) {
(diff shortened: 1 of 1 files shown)


gufengzhou:
Good one. The reason for the inefficiency was just we were trying to plot the transformation while creating it. After that we didn't clean it up. Thanks for the improvement!

PR #422, opened 2022-06-30
Clusters: potential improvement in Top Solution selection

andrey-legayev:
Hi,

I tried to understand how Robyn selects best model inside a cluster.
I've found some parts which could be fixed/improved - see below. 
Sorry in advance if I didn't understand the initial idea.

1. ```.clusters_df()``` in clusters.R sometimes returns ```Inf```
It happens as a result of dividing by zero which is unavoidable because of normalization in ```.min_max_norm()```
See screenshot of ```df``` and ```crit_df``` values inside ```.clusters_df()```:
image

2. These Inf values seem to be breaking error value calculation:
In general model with nrmse=8.19 (normalized and weighted) cannot have error=0, in my opinion it doesn't make sense
image

I prepared a fix with following changes:
 - switch weight formula from ```balance[1] / .data$nrmse``` to ```.data$nrmse / balance[1]``` to avoid ```Inf``` values
 - normalize weights to have 1 as a minimum value to see the same numbers before and after weight calculation if weights are ```c(1, 1, 1)``` (default value)
 - calculate error rate as ```^(1/2)``` instead of ```^-(1 / 2)``` - this simplification is possible after changing weighting formula
 - change final ranking sort accordingly to take best model

-- Andrey.

# Type of change

- Bug fix (non-breaking change which fixes an issue) and Code cleanup

# How Has This Been Tested?

Tested on our internal project.

### DIFF
--- a/R/R/clusters.R
+++ b/R/R/clusters.R
@@ -89,7 +89,7 @@ robyn_clusters <- function(input, all_media = NULL, k = "auto", limit = 1,
 
   # Select top models by minimum (weighted) distance to zero
   top_sols <- .clusters_df(cls$df, weights) %>%
-    mutate(error = (.data$nrmse^2 + .data$decomp.rssd^2 + .data$mape^2)^-(1 / 2)) %>%
+    mutate(error = (.data$nrmse^2 + .data$decomp.rssd^2 + .data$mape^2)^(1 / 2)) %>%
     .crit_proc(limit)
 
   output <- list(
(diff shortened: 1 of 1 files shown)


laresbernardo:
Hi @andrey-legayev these changes make sense to me. Thanks for the PR! I will check it on my end now.

andrey-legayev:
@laresbernardo  did you have a chance to check it?
I've just submitted another fix in calibration: #425

PR #464, opened 2022-08-15
update robyn_response() to receive numeric vector

Tomobay:
# Project Robyn
update robyn_response() to receive numeric vector.
current robyn_response allow to receive numeric value. If we want to get multiple result we needed to run multiple times or loop it. It takes so much time to try it. By allowing numeric vector it will reduce time of getting outputs.

Fixes # (issue)

# Type of change

- New feature (non-breaking change which adds functionality)

# How Has This Been Tested?
Unit test of robyn_response().
Tested 
- to allow numeric value and vectors.
- send error if it is not numeric vector.
- send error if it has negative value.

Checklist:

- Fork the repo and create your branch from master.
- If you've added code that should be tested, add tests.
- If you've changed APIs, update the documentation.
- Ensure the test suite passes.
- Make sure your code lints.

### DIFF
--- a/R/R/checks.R
+++ b/R/R/checks.R
@@ -723,19 +723,14 @@ check_allocator <- function(OutputCollect, select_model, paid_media_spends, scen
 
 check_metric_value <- function(metric_value, media_metric) {
   if (!is.null(metric_value)) {
-    if (length(metric_value) != 1) {
-      stop(sprintf(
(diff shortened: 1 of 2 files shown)


laresbernardo:
Thanks for this PR to improve `robyn_response()` function. Will merge now.

PR #498, opened 2022-09-21
Adding new optimization feature

Tomobay:
## Project Robyn
adding new feature to optimize based on historical spend by not changing total spend in that period.

## Type of change
- feat: New feature (non-breaking change which adds functionality)
Using daily, weekly, and monthly granularity data, the optimization is performed to set total cost used in the period are the same. If the cost of channel Facebook is set at 120%, the optimization method would set the cost  of channel Facebook for the entire period at 120% for each level of granularity. This is helpful in determining annual budgets and other long term budgets

## How Has This Been Tested?
- If you've changed docuemntation, run `devtools::document()`. Should update .Rd files.
- Make sure your code lints.

## Comments
Please review and tell me the function name  or place to marge to current function.
If you have any questions please let me know.


gufengzhou:
hi, thanks for submitting this. I see that you're adding a new dependency Rsolnp for the optimisation. Is there any reason you can't use the current one/ nloptr? Due to CRAN submission, we're conservative about adding new dependencies.

You've added the new parameter new_budget_ratio that scales the total spend, which is a good idea. But I do think it's a lot of new codes and functions for creating this in your solution, while theoretically you can just add the new multiplyer on histSpendUnitTotal. Or am I wrong about your intension?

Tomobay:
1. Let me modify to use nloptr.

2. It's different from current optimization logic. not using mean of spend but using actual historical spend itself.
I will explain it on Robyn hackathon .

PR #540, opened 2022-11-19
solution for preventing vectory memory exhaused

jbender5:
## Project Robyn

After upgrading to  Robyn v 3.8.1 I was experiencing vector memory exhausted error in which I could not export files from OutputCollect on my MacOS using 8k iterations and 4 trials.

Fixes # (issue)
In order to solve this, I increased the physical and virtual memory allocated through R_MAX_VSIZE=100Gb 

## Type of change
I added 4 lines of commented out code to increase memory

## How Has This Been Tested?
I am successfully running models that were previously causing the error

### DIFF
--- a/demo/demo.R
+++ b/demo/demo.R
@@ -343,6 +343,10 @@ print(OutputCollect)
 # pareto_media_transform_matrix.csv, all media transformation vectors
 # pareto_alldecomp_matrix.csv, all decomposition vectors of independent variables
 
+## MacOS Only: If vector memory exhaused, increase physical and virtual memory allocation
+# install.packages("usethis")
+# library(usethis)
+# usethis::edit_r_environ() # Once a new R studio tab opens, add this to the first line: R_MAX_VSIZE=100Gb
 
 ################################################################
 #### Step 4: Select and save the any model


laresbernardo:
Thanks for sharing @jbender5. I'm currently working on reducing the outputs size which will avoid these kinds of issues on memory crashes. A couple of questions:
- What's the difference on your approach vs setting `Sys.setenv('R_MAX_VSIZE'=32000000000)` manually without further dependencies (usethis package)? We can set these envionment settings with `Sys.setenv()` and then check with  `Sys.getenv("R_MAX_SIZE")` afterwards.
- Is it truly only for macOS users? What happens with other OS?

jbender5:
No problem. Reducing the outputs size is the best solution. Regarding your questions:

- Setting 'R_MAX_VSIZE' through Sys.setenv should have the same effect as updating the .Renviron directly because they are both updating the environmental variable; however, I still receive the vector memory exhausted error using the former method. Rather than adding a dependency on the 'usethis' package, you can use the command line directly to update .Renviron as outlined [here](https://stackoverflow.com/questions/51248293/error-vector-memory-exhausted-limit-reached-r-3-5-0-macos)
- Unfortunately, I only have have macOS for personal and professional use. I can't test on other OS

laresbernardo:
Hey @jbender5 I've just deployed [3.8.2](https://github.com/facebookexperimental/Robyn/releases/tag/v3.8.2) with a new way/order of calculating decomp vectors (which was the new super heavy list that was crashing some users' computation power) and I think this won't be needed for now. Thanks for providing a viable solution that worked for you though. I'll close this PR for now.

PR #600, opened 2023-01-10
feat: added ds for robyn_response()

laresbernardo:


R/R/response.R
```diff
@@ -245,24 +246,46 @@ robyn_response <- function(InputCollect = NULL,
...
+          # Set closest date available if not exact date provided
+          if (!metric_ds %in% get_ds) {
+            metric_ds_val <- get_ds[which(abs(get_ds - metric_ds) == min(abs(get_ds - metric_ds)))]
+            if (!quiet) warning("Input 'metric_ds' (", metric_ds, ") was not valid. Picking closest date: ", metric_ds_val)
+          } else {
+            metric_ds_val <- metric_ds
+          }
+          if (!quiet) message("Using ds '", metric_ds_val, "' for the adstocked input value")
+        } else {
+          ## Multiple metric_ds (and metric_value)
```
gufengzhou, line 269: this is the case when metric_ds has multiple value and metric_value has only 1, e.g. user inputs  "from-to" dates, and we just divide metric_value by the length of the from-to dates. this is the case when user has a budget for x days, but don't know how to split the spend for each day.

we can also allow users to input all desired periods, like for 3 days, 2023-01-01, 2023-01-02, 2023-01-03, while also inputting 3 metric_value like c(200, 150, 250). This is the case when the user has a future media plan and know when they want to spend how much. 

I agree that we can think of a more intuitive argument name than 'metric_ds'. we can consider it along the way.

R/R/transformation.R
```diff
@@ -147,7 +148,8 @@ adstock_weibull <- function(x, shape, scale, windlen = length(x), type = "cdf")
     x_decayed <- x
     thetaVecCum <- 1
   }
-  return(list(x = x, x_decayed = x_decayed, thetaVecCum = thetaVecCum))
+  inflation <- sum(x_decayed) / sum(x)
```
gufengzhou, line 151: It's good that we have this inflation output here. but this is just the total inflation, while for each date the inflation will be different. Now in robyn_response, we already calculate the inflation for those metric_ds dates, which will be used in the allocator later to scale up the raw spend for those specific periods. therefore, maybe rename the inflation here as inflation_total?

R/R/response.R
```diff
@@ -288,14 +288,8 @@ robyn_response <- function(InputCollect = NULL,
...
     media_vec_sim[metric_ds_loc] <- metric_value
-    if (adstock == "geometric") {
-      x_list_sim <- adstock_geometric(x = media_vec_sim, theta = theta)
-    } else if (adstock == "weibull_cdf") {
-      x_list_sim <- adstock_weibull(x = media_vec_sim, shape = shape, scale = scale, type = "cdf")
-    } else if (adstock == "weibull_pdf") {
-      x_list_sim <- adstock_weibull(x = media_vec_sim, shape = shape, scale = scale, type = "pdf")
-    }
-    m_adstocked_sim <- x_list_sim$x_decayed
+    m_adstocked <- transform_adstock(media_vec_sim, theta = theta, shape = shape, scale = scale)
```
gufengzhou, line 291: I like it that we functionalise this part. we do have this in various places through out the package, would be great to have them all changed to this. be very careful tho because it's very difficult to debug if this part is screwed.

R/R/response.R
laresbernardo: I think you missed the part of the code where this is already implemented haha. I've enabled the "last" or "last_n" inputs for "metric_ds" where n can be any integer value to pick from the latest n available days. The multiple metric_value is also enabled already and divided by total metric_ds dates. I still have to add "all" as a possible input though.

laresbernardo: Input "all" enabled. Please @gufengzhou can you check the outputs and see if they are as expected, especially for +1 ds. You can use these:
```
devtools::load_all()
robyn_response(
  InputCollect = InputCollect,
  OutputCollect = OutputCollect,
  select_model = select_model,
  media_metric = select_media,
  metric_value = metric_value,
  metric_ds = "last",
  quiet = FALSE
)
robyn_response(
  InputCollect = InputCollect,
  OutputCollect = OutputCollect,
  select_model = select_model,
  media_metric = select_media,
  metric_value = metric_value,
  metric_ds = "last_3",
  quiet = FALSE
)
robyn_response(
  InputCollect = InputCollect,
  OutputCollect = OutputCollect,
  select_model = select_model,
  media_metric = select_media,
  metric_value = metric_value,
  metric_ds = "all",
  quiet = FALSE
)
robyn_response(
  InputCollect = InputCollect,
  OutputCollect = OutputCollect,
  select_model = select_model,
  media_metric = select_media,
  metric_value = metric_value,
  metric_ds = c("2019-01-01", "2019-11-11"),
  quiet = FALSE
)
```

R/R/response.R
```diff
@@ -222,59 +214,206 @@ robyn_response <- function(InputCollect = NULL,
...
+            metric_value_updated <- rep(metric_value_updated, metric_ds_n)
+          }
+        } else {
+          ## Manually inputting each date
+          if (all(metric_ds %in% all_dates)) {
+            metric_ds_val <- metric_ds
+            metric_ds_loc <- which(metric_ds %in% all_dates)
+            avlb_dates <- all_dates[(all_dates >= min(metric_ds_val) & all_dates <= max(metric_ds_val))]
+            if (length(metric_ds_loc) != length(avlb_dates) & !quiet)
+              warning(sprintf("There are %s skipping dates within your 'metric_ds' input",
```
laresbernardo, line 390: Should we allow skipping dates to be able to plot saturation curves for a channel in a date range (that contained zeroes - which can be taken out manually too)?

```
library(patchwork)
channel <- "facebook_S"
metric_values <- InputCollect$dt_input[[channel]]
non_adst <- robyn_response(
  InputCollect = InputCollect,
  OutputCollect = OutputCollect,
  select_model = select_model,
  media_metric = channel,
  metric_value = metric_values[metric_values > 0],
  # metric_ds = "all",
  quiet = FALSE
)
yes_adst <- robyn_response(
  InputCollect = InputCollect,
  OutputCollect = OutputCollect,
  select_model = select_model,
  media_metric = channel,
  metric_value = metric_values[metric_values > 0],
  metric_ds = InputCollect$dt_input$DATE[metric_values > 0],
  quiet = FALSE
)
non_adst$plot + yes_adst$plot
```

PR #640, opened 2023-03-06
fix: inversed CPA in allocator viz table

laresbernardo:


laresbernardo:
LGTH @gufengzhou thanks for double checking!

PR #648, opened 2023-03-10
feat: allocator's new scenario for ROAS target

laresbernardo:


laresbernardo:
Error when one of the coef is zero: `Error in is.nloptr(ret) : x0 contains NA`

R/R/plots.R
```diff
@@ -862,24 +862,33 @@ allocation_plots <- function(InputCollect, OutputCollect, dt_optimOut, select_mo
       values = round(.data$values, 4),
       # Deal with extreme cases divided by almost 0
       values = ifelse((.data$values > 1e15 & .data$metric %in% c("ROAS", "mROAS")), 0, .data$values),
-      values_label = dplyr::case_when(
+      values_label = suppressWarnings(dplyr::case_when(
```
laresbernardo, line 865: Why `suppressWarnings()`? Should avoid these...

gufengzhou: yea it's a warning when having NAs. When there's a 0 coef channel, ROAS will be NA and the paste() function will throw warning of "coercing NA to character" for the whole labels, but the outcome is desired / correct. So I was lazy:) In case you want to solve that go ahead.

PR #683, opened 2023-04-12
recode: reduce the size of OutputModels (Part 1)

laresbernardo:


laresbernardo:
Using demo data, 1 trial and 1000 iterations:

```
#### Original (main)
> lares::what_size(OutputModels)
[1] "40 Mb"

#### With changes
> lares::what_size(OutputModels)
[1] "39.6 Mb"
```

PR #786, opened 2023-07-19
Bug fix for pareto_alldecomp_matrix.csv

arowley85:
## Project Robyn

This pull requests amends the creation of "pareto_alldecomp_matrix.csv".

Main currently produces what I believe are erroneous values against "rawSpend" and "predictedExposure" values in that csv. Regardless of input, the rawSpend and predictedExposure values appear to be always the same as "rawMedia" values (the latter being correct values). 

Having examined previous versions of the code I am reasonably confident that the intended behaviour is to carry the spend values from paid_media_spends as raw data directly into rawSpend. In addition, the predictedExposure values I believe are intended to correspond to the y_hat values which are originally created in fit_spend_exposure. 
 

## Type of change
- fix: Bug fix

## How Has This Been Tested?
I have tested the change by running the following scenarios:
- paid_media_vars is absent/NULL (in this case, behaviour is unchanged to current)
- paid_media_vars is present, no values are equal to paid_media_spend
- paid_media_vars is present, some values are equal to paid_media_spend

I have considered that in some cases yhat values are intended to come from the Michaelis–Menten procedure and in some cases the linear model depending if fit with MM was possible and the comparison of the R^2 values. I have relied on the population of values in InputCollect$modNLS$yhat where currently row values with models=='nls' are actually linear model values according to which model was chosen.

### DIFF
--- a/R/R/pareto.R
+++ b/R/R/pareto.R
@@ -345,14 +345,17 @@ robyn_pareto <- function(InputCollect, OutputModels,
       ## 4. Spend response curve
       dt_transformPlot <- select(dt_mod, .data$ds, all_of(InputCollect$all_media)) # independent variables
       dt_transformSpend <- cbind(dt_transformPlot[, "ds"], InputCollect$dt_input[, c(InputCollect$paid_media_spends)]) # spends of indep vars
-      dt_transformSpendMod <- dt_transformPlot[rw_start_loc:rw_end_loc, ]
-      # update non-spend variables
-      # if (length(InputCollect$exposure_vars) > 0) {
-      #   for (expo in InputCollect$exposure_vars) {
-      #     sel_nls <- ifelse(InputCollect$modNLSCollect[channel == expo, rsq_nls > rsq_lm], "nls", "lm")
-      #     dt_transformSpendMod[, (expo) := InputCollect$yhatNLSCollect[channel == expo & models == sel_nls, yhat]]
-      #   }
-      # }
+      dt_transformVars <- cbind(dt_transformPlot[, "ds"], InputCollect$dt_input[, c(InputCollect$paid_media_vars)])
+      dt_transformSpendMod <- tibble(ds = select(InputCollect$dt_mod,.data$ds))
+      for (i in seq_along(InputCollect$paid_media_vars)) {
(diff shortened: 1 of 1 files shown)


gufengzhou:
Hi, this is rather a legacy issue. Because we've deprecated modeling with exposure, the transformation now does not involve any exposure metrics anymore. I'd rather remove rawMedia and predictedExposure to reduce objective size. Are you using these actively?

arowley85:
Yes, I thought these metrics could still be used to assess the suitability of each channel and whether data needs to be changed. We do execute these transformations, so to not output anything, except some plots that have to be manually saved, seems a little clumsy. If this is not the right place, where is the right place?

gufengzhou:
The two categories can be removed because they're redundant. What's missing is the spend exposure fitting. I can expert the plots when available, does that help?

arowley85:
I already wrote a two liner to save down the plots. Something like this

`if(length(InputCollect$exposure_vars)>0){ pSpendExposure <- wrap_plots( InputCollect$modNLS$plots, ncol = ifelse(length(InputCollect$modNLS$plots) <= 3, length(InputCollect$modNLS$plots), 3) ) + plot_annotation( title = "Spend-exposure fitting with Michaelis-Menten model", theme = theme(plot.title = element_text(hjust = 0.5)) ) ggsave( paste0(OutputCollect$plot_folder, "spend_exposure_fitting.png"), plot = pSpendExposure, dpi = 600, width = 12, limitsize = FALSE, height = ceiling(length(InputCollect$modNLS$plots) / 3) * 7 ) }`

Even so, I think a csv would be more helpful. If it doesn't belong here, perhaps save down InputCollect$modNLS$yhat directly. Although, in that case, the user is left to rejoin the raw input values back to that in order to make any kind of comparison.

gufengzhou:
Thanks again for reporting this issue! As explained, this is a legacy issue due to the impression/exposure fitting. You're totally right about that: 

>  these metrics could still be used to assess the suitability of each channel and whether data needs to be changed

But this happens before the main iterations, while this csv is generated after the iterations. We'll fix this in the next commits by removing "predictedExposure" and "rawMedia" to avoid further confusion.

PR #826, opened 2023-09-29
Add objective weights

gufengzhou:
An experimental feature: adding an argument objective_weights to allow manual customization of objective function weightage. For example, with NRMSE & RSSD have same weights of c(1,1)  by default. Set to 0 is possible. WARNING: The weights cannot be optimised. Any customisation is arbitrary and has strong risk of bias.


igorskokan (approved):
LGTM

PR #830, opened 2023-10-04
Upgrade Docusaurus packages

zpao:
`yarn upgrade-interactive` and selected the docusaurus packages only. This obviously updated recursive dependencies too.

This seems to fix an issue with the mobile sidebar.


gufengzhou (approved):
LGTM and thank you so much!

PR #831, opened 2023-10-04
Fix React warnings/errors.

zpao:
This fixes warnings/errors that React is outputting.

1. `title` was an object, so the same `key` was being set for all elements. There's no need to use JSX for that since you're rendering an `<h3>` anyway, so easy win.
2. `<ul>` isn't a valid DOM child of `<p>` so rendering engines actually move DOM nodes around to make it work, which can break React. Use a `<div>` instead (and remove extra React fragment wrapping).
3. Properly case some attributes on `<iframe>`s

### DIFF
--- a/website/src/pages/index.js
+++ b/website/src/pages/index.js
@@ -17,66 +17,58 @@ import styles from './styles.module.css';
 
 const features = [
   {
-    title: <>Reduces human bias</>,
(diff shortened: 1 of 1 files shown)


gufengzhou (approved):
LGTM and thank you so much!

PR #843, opened 2023-10-19
feat: check and fix all_sol_json & new pareto_df parameters inputs

laresbernardo:
---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1205722685275404


gufengzhou (approved):
LGTM

PR #852, opened 2023-11-04
feat + fix: bundle fixes and improvements on json files

laresbernardo:
- **fix**: restructure and clear usage of `robyn_write()` for 3 use cases: inputs only, models and inputs, outputs, models and inputs, depending on users inputs. Last scenario is used to replicate specific models.
- **fix**: run timings on refresh models and robyn_run(... outputs = TRUE) now working
- **fix**: get rid of some ":="s in the code (data.tables dependency)
- **fix**: add "penalty" in HYPS_NAMES fixed vector


laresbernardo:
Good to land and merge @gufengzhou

gufengzhou (approved):
LGTM

PR #865, opened 2023-11-19
fix: compatibility with legacy files on robyn_recreate()

laresbernardo:
fix: compatibility with legacy files on robyn_recreate()


gufengzhou (approved):
LGTM

PR #868, opened 2023-11-23
fix: crashed cut() when monthly data

laresbernardo:
Issue with holidays data when monthly granularity fixed.
```
Error in mutate(., ds = cut(.data$ds, intervalType)) : 
  ℹ In argument: `ds = cut(.data$ds, intervalType)`.
Caused by error in `cut.default()`:
! 'x' must be numeric
```


gufengzhou (approved):
LGTM

PR #871, opened 2023-11-27
fix: when all values on a column are negative crash

laresbernardo:
When running `robyn_run()`, if a value contained "only negative values", it returned this error:

```
Error in { : 
  task 1 failed - "arguments imply differing number of rows: 16, 17"
```

Fixed by checking on `robyn_input()` when all values from any column are negative (or zero) and transform it with `abs()`.


gufengzhou (approved):
LGTM

Issue #872, opened 2023-11-29
Error in `mutate()`: ℹ In argument: `rsq = lares::winsorize(.data$rsq, thresh = c(0.01, 0.99))`


caarolinagil:
## Describe issue
I get a error when I try to run de robyn_fun() function. After running 5 trials and 2000 iterations with weibul_pdf, this message appear:

Error in `mutate()`:
ℹ In argument: `rsq = lares::winsorize(.data$rsq, thresh = c(0.01, 0.99))`.
Caused by error in `quantile.default()`:
! missing values and NaN's not allowed if 'na.rm' is FALSE

## Environment & Robyn version
Make sure you're using the latest Robyn version before you post an issue.
- Robyn 3.9.0
- R version 4.2.1

gufengzhou:
@laresbernardo would you please check this one? thanks!

laresbernardo:
It seems there is at least 1 value for R2 values that is missing (`NA`) there. Will add a quick fix by adding `na.rm = TRUE` to skip those, but probably a good idea to check why, how, and when this happens. Fix on the way...

caarolinagil:
Yes, I think it's a good idea to check that because it's a rare case.

Okay, is the fix done on any version of Robyn??

Thanks a lot!

laresbernardo:
@caarolinagil just landed in dev version v3.10.5.9008. Thanks for reporting it!

PR #875, opened 2023-11-30
fix: winsorize errors with na.rm = TRUE #872 + InputCollect print when no prophet

laresbernardo:
fix: winsorize errors with na.rm = TRUE #872
fix: InputCollect print when prophet is deactivated


gufengzhou:
looks good and pls add the version bump

laresbernardo:
Yup, done! Reado to approve and merge

PR #917, opened 2024-02-05
fix: intercept param in outputs + ts_validation plot for convergence

laresbernardo:
fix: intercept input not being stored (thus not exported in JSON file

fix: add ts_validation plot to all_plots and export regardless of ts_validation param for convergence viz


laresbernardo:
Example of (dummy) 1 trial and `ts_validation = FALSE` output generated
da68b514-43a0-47ed-aae8-4f836574c6c0

gufengzhou (approved):
lgtm

PR #928, opened 2024-02-19
feat: add clusters information to one-pagers, fix metric notes, reduce texts sizes, no ng check to recreate + fix: RNG warning

laresbernardo:


gufengzhou (approved):
Lgtm

PR #929, opened 2024-02-20
docs: clarify objective_weights order

laresbernardo:


gufengzhou (commented):
Bit confusing to have example of c(2,1) for only two objs, but there're 3 in total. We should explain when not calibrating, only RMSE and Rssd, then c(2,1) gives double weight to the first etc. Just more clarity please

laresbernardo:
Alright, clarified.

PR #930, opened 2024-02-23
feat: keep custom exported data from JSON when recreating model & use raw_data if available + update docs

laresbernardo:


gufengzhou (approved):
Lgtm

Issue #946, opened 2024-03-28
Cannot plot model onepager with ts_validation = False with "non-numeric argument to mathematical function" error


rupatrawut:
## Describe issue
If I train Robyn model with `ts_validation` set to `False`, everything looks fine in training phrase. I got all the output such as pareto_aggregated.csv, etc without problem. 

However, when calling `robyn_outputs` api via plumber, I got error message in robyn_log file as below:

> >> Plotting 20 selected models on 7 cores...
> |                                                                            
> |                                                                      |   0%Failed exporting results, but returned model results anyways:
> Error in {: task 1 failed - "non-numeric argument to mathematical function"

I also try this function but return following error.
```
#write one pager to specified location
load_onepager(sol='1_5_1', write=True, InputJson=InputCollect, OutputJson=OutputCollect, path=output_folder)
```

> ----> [4](vscode-notebook-cell:?execution_count=37&line=4) load_onepager(sol='1_5_1', write=True, InputJson=InputCollect, OutputJson=OutputCollect, path=output_folder)
> 
> File [c:\Users\rupatrawut\workspace\ml-marketing-mix-modeling\ml_mmm_robyn\robyn_api\python_helper.py:263](file:///C:/Users/rupatrawut/workspace/ml-marketing-mix-modeling/ml_mmm_robyn/robyn_api/python_helper.py:263), in load_onepager(InputJson, OutputJson, path, sol, top_pareto, write, max_size)
>     [261](file:///C:/Users/rupatrawut/workspace/ml-marketing-mix-modeling/ml_mmm_robyn/robyn_api/python_helper.py:261) sol_id = sol
>     [262](file:///C:/Users/rupatrawut/workspace/ml-marketing-mix-modeling/ml_mmm_robyn/robyn_api/python_helper.py:262) onepager = load_modeldata(sol_id,InputJson=InputJson,OutputJson=OutputJson)
> --> [263](file:///C:/Users/rupatrawut/workspace/ml-marketing-mix-modeling/ml_mmm_robyn/robyn_api/python_helper.py:263) image_data = binascii.unhexlify("".join(onepager))
>     [264](file:///C:/Users/rupatrawut/workspace/ml-marketing-mix-modeling/ml_mmm_robyn/robyn_api/python_helper.py:264) if(write==True):
[...]

PR #965, opened 2024-05-04
Fixed error when ts_validation is false.

yu-ya-tanaka:
## Project Robyn

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context.

Fixes # (issue)
Fixed error when ts_validation is false. When ts_validation is false, one-pager was not generated.

## Type of change
- fix: Added "@serializer json list(digits = 20, na = 'null')" for all post methods. This decorator prevents 1. rounding of a decimal, and 2. deletion of columns where the value are NA in the response of API.
- fix: Updated the function to transform json of OutputCollect to R object.
- update: Removed the warning message when converting the ggplot objects into hex, by adding suppressMessages(). 

## How Has This Been Tested?
- Tested on sample notebook and confirmed works correlty.


yu-ya-tanaka:
This RP is to solve this issue. https://github.com/facebookexperimental/Robyn/issues/946

gufengzhou (approved):
LGTM

PR #969, opened 2024-05-06
bl01: refresh fixes and improvements + others

laresbernardo:
After testing with few beta users on bl01 branch, it's good to land. 

- fix: several bugs when refreshing with `robyn_refresh()` - non-existing folders, cluster = FALSE, CSV exports, ts_validation, window_end inherited
- fix: penalties hyperparameters checked correctly when recreating a model that used penalization
- fix: similar plotting scales for refresh plots when performance is too large to improve readability (`refresh_plots_json()`)
- feat: new `add_data` parameter in `robyn_write()` to include raw data. When used, `dt_input` is not required to recreate model
- feat: improved chain logic on refresh models. If user is not following chained directories but models are available in `plot_folder`, they will be loaded and used
- feat: recreate InputCollect when RobynModel-models.json provided (no selected model)

This is still version 3.10.7 but will be CRAN's 3.11.0 if no bugs or issues are reported within a week or so after it lands (@gufengzhou)


gufengzhou (approved):
LGTM

Issue #960, opened 2024-04-25
Robyn Refresh error


bart-vanvlerken:
Hi all, hope you're doing well. I've found a bug in the latest github version of Robyn where I can't refresh my model. In addition, the output tells me 'Provided train_size but ts_validation = FALSE. Time series validation inactive.' when I set ts_validation = TRUE in my base model. Hope anyone can help!

![image](https://github.com/facebookexperimental/Robyn/assets/118725564/d3cb1fd5-6640-4451-adb5-28bd29915ccb)

![image](https://github.com/facebookexperimental/Robyn/assets/118725564/26f4eb3b-94fc-4368-b69a-a6e1b0ca7bee)

shuvayan:
I am also getting the same error while refreshing the model :
![image](https://github.com/facebookexperimental/Robyn/assets/11851253/94730f12-097a-4005-ad22-b036f0a2a8b9)

I think the issue is with the robyn_chain function as the "json_new$InputCollect$refreshSourceID" part returns NULL.
Maybe it is not being able to read the correct file.

laresbernardo:
There are a couple of bugs I've found in refresh functionality and am working on fixes on my end. Can you try updating Robyn with `Robyn::robyn_update(ref = "bl01")` and see how it goes? If it goes well I can merge to main afterwards.

laresbernardo:
@shuvayan @bart-vanvlerken we are planning on landing a CRAN stable version this week. Please, let us know asap if the issue persists with the version of the branch I shared so we can move forward or hold it.

bart-vanvlerken:
@laresbernardo @gufengzhou I'm going to try Bernardo's latest version today and will let you know! Thanks for all the help.

bart-vanvlerken:
Update. I updated to Bernardo's latest version, and the refresh functionality works again.
![image](https://github.com/facebookexperimental/Robyn/assets/118725564/70e1ab95-80bd-4c98-bbae-635bfab685cd)

However, I'm running into the same error: 'Provided train_size but ts_validation = FALSE. Time series validation inactive.' when I set ts_validation = TRUE in my base model. In addition, I'm getting two new peculiar warning messages (3 & 4, see screenshot below). 
![image](https://github.com/facebookexperimental/Robyn/assets/118725564/7fd4552b-4f71-4af0-b2f0-5e98ef8c7570)

[...]

PR #974, opened 2024-05-13
fix: refresh hyps check + use data available in json + refresh hyps + upper constraints fix when higher than mean

laresbernardo:
Fix on #960


gufengzhou:
Thanks for updating @laresbernardo . Would you please test the current change vs the main branch on refresh each to see if everything is identical? It'd be identical when you compare the JSONs and the list "hyper_values" are having the same values. Also when testing refresh, make sure to test 2 separate refresh steps (rf1 & rf2) to see if the whole chaining etc work properly. Thanks!

gufengzhou:
Thanks for digging. Theoretically refresh should also iterate through penalty. we just didn't have users going this far so far.

laresbernardo:
Alright, did some tweaks all around and I feel more confident on refreshing results. I've:
- feat: created a new `bounds_freedom` parameter so that users can overwrite the flexibility on hyperparameters bounds for refreshed models (default remains the same)
- feat: more flexibility on chain logic so original file may be elsewhere
- fix: fixed hyperparameters for lambda, penalties, train_test are now flexible, using the refresh flexibility bounds. Results are way more consistent and similar to original model
- fix: issues with refresh plots when chain is not followed correctly

laresbernardo:
@gufengzhou can you please check the [Fitted vs actual calculations](https://github.com/facebookexperimental/Robyn/blob/fb2179535c5fe84c14be7b4df35db11fd392d019/R/R/pareto.R#L459-L493) when refreshing? For some reason, the baseline is a large number and the predictions are small numbers (as if they haven't been scaled); so you get almost a straight line. Funny thing is that we haven't specifically changed anything there.

amanrai2508:
Hi @laresbernardo 

Can you check this issue as well , I am getting way different result when I am doing data refresh (for 2 weeks of data)
 : https://github.com/facebookexperimental/Robyn/issues/985

gufengzhou (approved):

gufengzhou:
> @gufengzhou can you please check the [Fitted vs actual calculations](https://github.com/facebookexperimental/Robyn/blob/fb2179535c5fe84c14be7b4df35db11fd392d019/R/R/pareto.R#L459-L493) when refreshing? For some reason, the baseline is a large number and the predictions are small numbers (as if they haven't been scaled - check rowSums(xDecompVec)); so you get almost a straight line. Funny thing is that we haven't specifically changed anything there... unless I'm getting terrible fit because of the hyps constraints(?)

If you check report_aggregated.csv, you'll see that the intercept are wrong. It was dropped in the initial model but not dropped in the refresh. The dropping of intercept happens in the refit function within robyn_mmm @laresbernardo

amanrai2508:
Hi @laresbernardo @gufengzhou 
This is the same issue that I am facing as well : https://github.com/facebookexperimental/Robyn/issues/985
If you need any help with respect to testing, I can help you with this. (Robyn refresh is flawed currently because of this)

> @gufengzhou can you please check the [Fitted vs actual calculations](https://github.com/facebookexperimental/Robyn/blob/fb2179535c5fe84c14be7b4df35db11fd392d019/R/R/pareto.R#L459-L493) when refreshing? For some reason, the baseline is a large number and the predictions are small numbers (as if they haven't been scaled - check rowSums(xDecompVec)); so you get almost a straight line. Funny thing is that we haven't specifically changed anything there... unless I'm getting terrible fit because of the hyps constraints(?)

laresbernardo:
Ok. It looks good now. @amanrai2508 would you mind updating Robyn to this branch (`Robyn::robyn_update(ref = "bl02")` + refresh R session) and testing the refresh functionality? Before we merge I'd like to hear from you. Thanks!

amanrai2508:
Hi @laresbernardo 

This is currently showing when I am trying to refresh : Error in check_nas(dt_input) : could not find function "check_nas"

amanrai2508:
Hi @laresbernardo 
Now the refresh looks good. (For some data it is off but for most of the data it is working)
Just wanted to know why we are plotting onepager of a solutionid where topsolution is False ( Though the results are similar in this solution as well)
image

image

laresbernardo:
Could you please check if model 1_142_5 is the solution with the lowest DECOMP.RSSD error across all Pareto-front models? The top solutions are the minimum combined error models per cluster, which may not match.

amanrai2508:
Yeah, it has the lowest DECOMP.RSSD. 
It looks fine on my end @laresbernardo ,now the values are more credible.

PR #1011, opened 2024-06-27
fix: one-pager decomp labels and warning

laresbernardo:
Checking with CRAN if this one can land ASAP.


laresbernardo:
> package Robyn_3.11.1.tar.gz is on its way to CRAN.

gufengzhou (approved):
lgtm

Issue #1003, opened 2024-06-20
Hyperparameters out of bounds


malaman5:
## Describe issue
Im getting this error when running robyn_inputs()

Error in check_hyper_limits(hyperparameters_ordered, "thetas") : 
  pinterest_spend_thetas's hyperparameter must have upper bound <1

The hyperparameter upper bound for  is currently set to <1 in the json file:
"pinterest_spend_alphas": [
        0.5,
        3
      ],
      "pinterest_spend_gammas": [
        0.3,
        1
      ],
      "pinterest_spend_thetas": [
        0.1,
        0.4
      ],

## Provide reproducible example
input_collect <- robyn_inputs(
  dt_holidays = dt_prophet_holidays,
  dt_input = dataset,
  json_file = parameters_file
)

The hyperparameters have been included in the parameters_file as a JSON just like other networks. When checking the input_collect object, the hyperparameters DO NOT include the pinterest hyperparameters. Not sure if it's not included due to the error or the error comes because the hyperparameters are not correctly declared. The pinterest hyperparameters have the very same structure as any other network which is parsed completely fine.

## Environment & Robyn version
- ‘3.11.0’ (also occurred with 3.10.3)
- "R version 4.3.2 (2023-10-31)"

PR #1013, opened 2024-07-03
fix: skip other hyps from old flow to replace exposure hyps with spend hyps #1003 + new exported functions

laresbernardo:
After checking #1003 with the user's data and script, there were 2 issues (not sure if not correlated):

1. The order of the exposure variables didn't match the order of the paid media spends variables + they used the same names. 
2. The old flow code didn't get rid of the HYPS_OTHERS (which include lambda) and when sorting all the hyps, the index was shifted.

Given the user is running Robyn in a pipeline and exposure metrics are only used to provide recommendations in console to provide a narrower granularity, I've suggested the user to simply set the exposure variables tu NULL.


laresbernardo:
@gufengzhou additionally to the fix, I've just exported some previously unexported functions to facilitate some calculations / standardize some code, especially on the immediate vs carryover calculations which can now be used as a side function and can calculate for specific date ranges. I've tested it and there are no changes on final users' end nor results. It's OK to land.

gufengzhou (approved):
LGTM

PR #1043, opened 2024-09-03
Implementation of some mmmdata_validation

shivkanthb:
Project Robyn
-------------

This commit implements several validation methods in the MMMDataValidation class within mmmdata_validation.py. These methods are Python equivalents of the validation checks performed in the R version of Robyn, specifically in the checks.R file.

The implemented methods include:

-   check_missing_and_infinite: Checks for missing (NA) and infinite values in the input data.

-   check_no_variance: Identifies columns with no variance in the input data.

-   check_date_variable: Validates the date variable in the input data.

-   check_dependent_variables: Ensures the dependent variables are valid and of the correct type.

These implementations aim to maintain parity with the R version while adapting to Python's syntax and pandas DataFrame operations.

Fixes # (issue number, if applicable)

Type of change
--------------

-   feat: New feature (non-breaking change which adds functionality)

How Has This Been Tested?
-------------------------

-   The code has been tested locally to ensure it runs without errors.

-   Unit tests for these methods will be added in a subsequent commit.

-   The implementations have been reviewed to ensure they match the functionality of the R version.

Note: This is part of an ongoing effort to port Robyn from R to Python. Additional validation methods will be implemented and committed based on feedback from this initial set of implementations.

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -6,44 +6,42 @@ name: Robyn Python application
 
 on:
   push:
-    branches: [ "robynpy_release" ]
+    branches: ['robynpy_release']
   pull_request:
-    branches: [ "robynpy_release" ]
+    branches: ['robynpy_release']
 
 permissions:
   contents: read
 
 jobs:
   build:
-
     runs-on: ubuntu-latest
 
     strategy:
-          matrix:
-            python-version: ["pypy3.9", "3.9", "3.10"]
-        
+      matrix:
+        python-version: ['pypy3.9', '3.9', '3.10']
+
     steps:
-    - uses: actions/checkout@v4
-    - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v5
-      with:
-        python-version: ${{ matrix.python-version }}
-        cache: 'pip'
-    - name: Display Python version
-      run: python -c "import sys; print(sys.version)"
-        
-    - name: Install dependencies
-      run: |
-        python -m pip install --upgrade pip
-        pip install flake8 pytest pytest-cov
-        if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
-    - name: Lint with flake8
-      run: |
-        # stop the build if there are Python syntax errors or undefined names
-        flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics --exclude=./robyn_api/*.py
-        # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
-        flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics --exclude=./robyn_api/*.py
-    # - name: Test with pytest. Enable this once first set of tests are written
-    #   run: |
-# 
-#        pytest ./python/tests --doctest-modules --junitxml=junit/test-results.xml --cov=com --cov-report=xml --cov-report=html
+      - uses: actions/checkout@v4
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+          cache: 'pip'
+      - name: Display Python version
+        run: python -c "import sys; print(sys.version)"
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install flake8 pytest pytest-cov
+          if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+      - name: Lint with flake8
+        run: |
+          # stop the build if there are Python syntax errors or undefined names
+          flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics --exclude=./robyn_api/*.py
+          # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
+          flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics --exclude=./robyn_api/*.py
+      - name: Test with pytest. Enable this once first set of tests are written
+        run: |
+          pytest ./python/tests --doctest-modules --junitxml=junit/test-results.xml --cov=com --cov-report=xml --cov-report=html
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -19,7 +19,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: ['pypy3.9', '3.9', '3.10']
+        python-version: ['3.9', '3.10']
 
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -31,6 +31,10 @@ jobs:
       - name: Display Python version
         run: python -c "import sys; print(sys.version)"
 
+      - name: updating PATH to enable importing robyn modules
+        run: |
+          echo "export PYTHONPATH=$PYTHONPATH:$(pwd)/python" >> $GITHUB_ENV
+
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -33,7 +33,7 @@ jobs:
 
       - name: updating PATH to enable importing robyn modules
         run: |
-          echo "export PYTHONPATH=$PYTHONPATH:$(pwd)/python" >> $GITHUB_ENV
+          echo "PYTHONPATH=$PYTHONPATH:$(pwd)/python/src" >> $GITHUB_ENV
 
       - name: Install dependencies
         run: |
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -35,7 +35,7 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install flake8 pytest pytest-cov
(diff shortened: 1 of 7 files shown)


python/src/robyn/data/validation/mmmdata_validation.py
```diff
@@ -15,46 +17,176 @@ def check_missing_and_infinite(self) -> ValidationResult:
...
+            mmmdata_spec.date_var,
+            mmmdata_spec.context_vars,
+            mmmdata_spec.paid_media_spends,
+            mmmdata_spec.organic_vars
+        ]
+        vars_to_check = [var for var in vars_to_check if var is not None]
+        vars_to_check = [item for sublist in vars_to_check for item in (sublist if isinstance(sublist, list) else [sublist])]
+
+        # Check for duplicates
+        duplicates = [var for var in set(vars_to_check) if vars_to_check.count(var) > 1]
```
sumane81, line 91: This checks duplicates across all variables vs duplicates within the set. do we know if a column can be dep_var as well as paid_media_spends etc? if that's the case, duplicates needs to be checked only within set.

python/src/robyn/data/validation/calibration_input_validation.py
```diff
@@ -3,7 +3,7 @@
 from robyn.data.validation.validation import Validation, ValidationResult
 from typing import List
 
-from robyn.modeling.entities.modelrun_trials_config import TrialsConfig
+# from robyn.modeling.entities.modelrun_trials_config import TrialsConfig
```
sumane81, line 6: let's just remove this line

python/src/robyn/data/validation/calibration_input_validation.py
```diff
@@ -30,7 +30,7 @@ def check_obj_weight(
     def check_iteration(
         self,
         calibration_input: CalibrationInput,
-        trials_config: TrialsConfig,
+        # trials_config: TrialsConfig,
```
sumane81, line 33: Let's remove this line as well

python/src/robyn/data/validation/mmmdata_validation.py
```diff
@@ -15,46 +17,176 @@ def check_missing_and_infinite(self) -> ValidationResult:
...
     def check_date_variable(self) -> ValidationResult:
         """
         Checks if the date variable is correct.
         
         :return: True if the date variable is valid, False otherwise.
         """
-        raise NotImplementedError("Not yet implemented")
+        mmmdata_spec = self.mmm_data.mmmdata_spec
+        dt_input = self.mmm_data.data
+        date_var = mmmdata_spec.date_var
```
sumane81, line 123: default value for date_var is "auto". Either define what "auto" means and update code or just remove "auto" as default value so that it becomes mandatory for user to specify the date variable name.

python/src/robyn/data/validation/mmmdata_validation.py
```diff
@@ -15,46 +17,176 @@ def check_missing_and_infinite(self) -> ValidationResult:
...
+            mmmdata_spec.organic_vars
+        ]
+        vars_to_check = [var for var in vars_to_check if var is not None]
+        vars_to_check = [item for sublist in vars_to_check for item in (sublist if isinstance(sublist, list) else [sublist])]
+
+        # Check for duplicates
+        duplicates = [var for var in set(vars_to_check) if vars_to_check.count(var) > 1]
+
+        # Check for invalid characters (spaces)
+        invalid = [var for var in vars_to_check if ' ' in var]
```
sumane81, line 94: What is the definition of invalid here? (do we need regex to cover more invalid cases?)

python/src/robyn/data/validation/mmmdata_validation.py
```diff
@@ -15,46 +17,176 @@ def check_missing_and_infinite(self) -> ValidationResult:
...
+        
+        error_details = {}
+        error_message = ""
+        
+        if dep_var not in dt_input.columns:
+            error_message = f"Dependent variable '{dep_var}' not found in the input data."
+        else:
+            if not pd.api.types.is_numeric_dtype(dt_input[dep_var]):
+                error_message = f"Dependent variable '{dep_var}' must be numeric."
+            elif dep_var_type not in ["revenue", "conversion"]:
```
sumane81, line 167: This is enum. No validation required for dep_var_type

python/src/robyn/data/validation/calibration_input_validation.py
sumane81, line 1: can you make sure it is #pyre-strict?

sumane81 (changes requested):
Review comments added.

python/src/robyn/data/validation/mmmdata_validation.py
shivkanthb: This is the only check R file does for invalid names

python/src/robyn/data/validation/mmmdata_validation.py
shivkanthb: In the R implementation, the function checks for duplicates both within each set of variables and across all variables combined. If any duplicates or missing variables are found, it raises an error with a specific message indicating the problem.

python/src/robyn/data/validation/mmmdata_validation.py
shivkanthb: Good call. In R, the code automatically checks for a column that matches is.Date. In python Ill have throw an  error for now.

python/src/robyn/data/validation/mmmdata_validation.py
sumane81: Let's think beyond current R code...

    invalid_cols = []
    for col in df.columns:
        if not isinstance(col, str):
            invalid_cols.append(col)
        elif not col.isidentifier():
            invalid_cols.append(col)
    return invalid_cols
  
    
    # Example usage
df = pd.DataFrame({
    'valid_col': [1, 2, 3],
    'invalid-col': [4, 5, 6],  # invalid due to hyphen
    '123invalid': [7, 8, 9],  # invalid due to starting with digit
    ('tuple', 'col'): [10, 11, 12]  # invalid due to being a tuple
})

shivkanthb: Are these rules enforced somewhere? Why cant we have columns with hypen?

sumane81: I am rethinking.. It doesn't matter if it starts with number or has hypen. 1. it should be valid string. 2. " " is also valid string so need to trim it 3. leading and trailing spaces also need be trimmed before checking if it is string.

shivkanthb:
Screenshot 2024-09-04 at 4 49 57 PM

python/src/tests/test_mmmdata_validation.py
```diff
@@ -0,0 +1,111 @@
...
+
+from robyn.data.entities.mmmdata import MMMData
+from robyn.data.validation.mmmdata_validation import MMMDataValidation
+
+
+class TestMMMDataValidation(unittest.TestCase):
+
+    def setUp(self):
+        # Create a sample MMMData object for testing
+        data = pd.DataFrame({
```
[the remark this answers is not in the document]
shivkanthb: Yep already tested with demo data in tutorials and worked fine. Added new data here to check more broadly

python/src/tests/test_mmmdata_validation.py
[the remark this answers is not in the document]
shivkanthb: done

python/src/tests/test_mmmdata_validation.py
```diff
@@ -0,0 +1,111 @@
...
+        result = self.validation.check_dependent_variables()
+        self.assertFalse(result.status)
+        self.assertIn('dependent_variable', result.error_details)
+
+    def test_validate(self):
+        results = self.validation.validate()
+        self.assertEqual(len(results), 5)  # Ensure all 5 validations are performed
+        self.assertTrue(all(result.status for result in results))
+
+if __name__ == '__main__':
```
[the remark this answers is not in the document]
shivkanthb: it wont be executable now. But also didnt need the @pytest annotation for it to work.

.github/workflows/python-app.yml
```diff
@@ -6,44 +6,42 @@ name: Robyn Python application
...
+          python-version: ${{ matrix.python-version }}
+          cache: 'pip'
+      - name: Display Python version
+        run: python -c "import sys; print(sys.version)"
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install flake8 pytest pytest-cov
+          if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
```
sumane81, line 38: Update requirements.txt to ./python/requirements.txt or python/requirements.txt and see if that fixes the dependencies issue.

shivkanthb: updated

.github/workflows/python-app.yml
```diff
@@ -6,44 +6,46 @@ name: Robyn Python application
...
   build:
-
     runs-on: ubuntu-latest
 
     strategy:
-          matrix:
-            python-version: ["pypy3.9", "3.9", "3.10"]
-        
+      matrix:
+        python-version: ['pypy3.9', '3.9', '3.10']
```
sumane81, line 22: github action build for pypy3.9 fails. There is specific reason we put this here so try updating it to pypy3.10 if it solves the issue.

shivkanthb: deleted for now

sumane81 (approved):

PR #1044, opened 2024-09-05
Adding python logger config

sumane81:
## Project Robyn

Added Python logger config.
1. robyn/common/config/logger.conf contains logger configuration. Currently it is set to print INFO for both file and console handler
2. log file will be created in current working directory of end usage. for example - log file will be created in tutorials directory if any tutorials is run. (not in working directory specified with instantiation of Robyn)
3. CalibrationInputValidation was broken and it was blocker to test logger changes so fixing it here as well.

## How Has This Been Tested?
image

image


alxlyj (approved):
Looks good to me.

shivkanthb (approved):

PR #1061, opened 2024-09-18
feat: cleaner decomp_plot() using cols instead of areas + ref line in 0

laresbernardo:
To avoid white spaced and weird plots with spaces when channels impact changed to negative in some periods, migrating plot to columns instead of areas in `decomp_plot()`. Also, added ref line at zero.

See previous and new looks attached for reference. 
new
old

Complexity implementing: date is a continuous value; for columns to have no space between columns, they must be discrete. Solved.

### DIFF
--- a/R/DESCRIPTION
+++ b/R/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: Robyn
 Type: Package
 Title: Semi-Automated Marketing Mix Modeling (MMM) from Meta Marketing Science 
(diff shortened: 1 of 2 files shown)


laresbernardo:
Also got rid of no-impact channels from plot/legend and sorted by absolute impact (as in lots of negative and lots of positive will be prioritized compared with lots of positive only - same dims). See below.

new2

gufengzhou (approved):
LGTM

PR #1062, opened 2024-09-18
visualizer interface

shivkanthb:
## Project Robyn

This PR adds a new visualization module to the Python port of Robyn. The change introduces a `RobynVisualizer` class with methods that correspond to the various plotting functions used in the R version of Robyn. This new module will enable the Python version to generate similar visualizations as the R version, enhancing the functionality and user experience of the Python port.

Fixes #[Issue number for Python port visualization]

## Type of change
- feat: New feature (non-breaking change which adds functionality)

## How Has This Been Tested?
- Fork the repo and create your branch from master (latest dev version).
- The new `RobynVisualizer` class has been added with method declarations that mirror the R version's plotting functions.
- No actual implementation has been added yet, so full testing will be required once the methods are implemented.
- Ensure all existing tests pass after adding this new module.
- Make sure the new code follows the project's coding standards and lints properly.

Note: This PR only includes the structure for the visualization module. Actual implementation of the plotting functions will be needed in subsequent PRs. Documentation updates will also be required once the implementation is complete.


shivkanthb:
This is not fully ready it but more of a v0 scaffold

python/src/robyn/visualization/visualizer.py
sumane81, line 1: Is this single visualizer class for all the plots?
As discussed, we should be following single responsibility principle here. all plots plotted using input data vs modeling vs allocator etc should reside in it's own classes with relevant data received in init method.

sumane81 (changes requested):
module needs to broken down into smaller, relevant plotters

shivkanthb:
Added different visualizer classes based on all the plots in demo.R. This is a good start and we can add more plot implementations as we encounter them. 

Here's a sample oh how you would call a particular plot with the above interfaces

To call a specific plot like plot_allocator, you would use the RobynVisualizer class. 

1. First, create an instance of the RobynVisualizer class:
```
visualizer = RobynVisualizer()
```

2. Then, you would call the plot_allocator method, passing in the necessary data:
```
allocator_collect = {...}  # This would be your allocator data
fig = visualizer.plot_allocator(allocator_collect)
```
The fig variable now contains the matplotlib Figure object, which you can display or save as needed.

shivkanthb:
the BaseVisualizer would look something like
```
class BaseVisualizer:
    def __init__(self):
        pass

    def _setup_plot(self):
        # Common plot setup logic
        fig, ax = plt.subplots(figsize=(10, 6))
        ax.set_xlabel('X-axis label')
        ax.set_ylabel('Y-axis label')
        ax.grid(True)
        return fig, ax
 ```

**Comment by shivkanthb** [comment]
Update:
To display a plot, you could

```
visualizer = RobynVisualizer()
# ... set up visualizer ...
fig = visualizer.plot_adstock()
plt.show()  # This will display the plo
```

dhavalpatel624624 (approved):

PR #1075, opened 2024-10-14
Updating pareto-optimizer until it generates plot data

dhavalpatel624624:
## Project Robyn
##Pareto Optimizer Feature without the data optimizer
Updated data mapper to map outputmodel from R to python
Calculates pareto front
This PR adds the logic that rewrite response.R based on the call from Pareto_optimizer
It also Calculates the hill hyperparameters from the response results.

## How Has This Been Tested?
Tested using test data from the `robyn_python_notebook.ipynb` before it calls robyn_output api endpoint.
Generated this new notebook to test the python logic
image


shivkanthb:
good stuff

PR #1079, opened 2024-10-21
Fix: fix hyperparameter definitions for modeling, update to logger

alxlyj:
## Project Robyn

As titled. In this PR, we are:

- Fixing print calls to use logger.
- Fixing Hyperparameter definition. Rather than splitting via string, we are correctly calling and using the data classes used in hyperparamters.py. 
- Regenerated the notebooks (tutorial2, tutorial3) to use the right calls. 
- Change inconsistent datatype for train_size. Changed to List. 

### Test plan

image

#### Feature engineering
image

#### Modeling
image


sumane81 (approved):
Looks good to me.

PR #1080, opened 2024-10-22
Robynpy pareto plot data impl

dhavalpatel624624:
## Project Robyn
This PR adds implementation of transformation module functions.
Fixes the implementation of response_curve with integration with transformation.
Completes the implementation of generate_plot function of pareto_optmizer

Fixes # (issue) 
datamapper, and existing notebook for running pareto optimizer e2e for generating the plot data from mmmdata and outputmodel

## Type of change
Implements the core functionality of pareto_optimizer which generates the plot data which will be later on used for clustering and plotting.

## How Has This Been Tested?
Using the test data to run the e2e workflow, exported the data right before the call to pareto optimizer into a json file and read that in using datamapper. Used the outputs from the datamapper to then call pareto_optimizer e2e and validated results with R.
R Output:
image

Python Output:
image


python/src/tutorials/tutorial4_pareto.ipynb
sumane81, line 1: tutorial refers to the test_data.json but it's not in the repo or part of this PR. I couldn't run the tutorial

dhavalpatel624624: these test_data files are too large for git (>100mb) I can send you the link to it to download and try it locally

sumane81 (approved):
Too big diff to review. Looking at the test plan, approving it for now to continue with implementation of next components.

PR #1096, opened 2024-10-29
One pager interface

sumalreddy17:
Add One Pager Plot class structure and method signatures

Introduces core class architecture for marketing analytics visualization
with comprehensive method signatures for all plot types.

Key components:
- PlotData dataclass for standardized data handling
- MarketingOnePager class with plot generation methods
- Method signatures for 8 distinct visualization types


dhavalpatel624624:
Let's avoid the use of both pyplot and go. Lets pick one over the other. Also, I see some files have Any datatypes, lets get those specified. Can we also add the  `PlotData`  in the constructor of each individual visualizer files and remove it from underlying methods if it makes sense. Then, make sure you instantiate each of the visualizers and call the functions in `generate_onepager` instead of the functions directly.

python/src/robyn/visualization/plot_data.py
sumane81, line 1: You shouldn't need to define new data entity in visualizer module. Input data for visualizer should directly come from component. For example - ParetoVisualizer should take ParetoResult.

python/src/robyn/visualization/onepager_visualizer.py
```diff
@@ -0,0 +1,116 @@
...
+    ):
+        """
+        Initialize OnePager with required results objects.
+        
+        Args:
+            pareto_result: Results from Pareto optimization
+            clustered_result: Optional clustering results
+            adstock: Optional adstock configuration
+        """
+        self.pareto_result = pareto_result
```
sumalreddy17, line 38: This can be moved to BaseVisualizer.

sumane81 (changes requested):
Comments added

dhavalpatel624624 (approved):

dhavalpatel624624:
Looks good, thanks for making all the changes! :)

python/src/robyn/visualization/input_visualizer.py
sumane81, line 1: input visualizer is vague name. Can we make it more specific?

sumalreddy17: How about media_response_visualizer?

python/src/robyn/visualization/robyn_visualizer.py
sumane81, line 1: What is this file for? Can you delete it if not needed?

sumalreddy17: Offline decision to clean up in a different PR

sumane81 (approved):
Some clean up is needed but that will come in separate PR.

PR #1101, opened 2024-10-30
Test/unit tests for response curve

Marco-Premier:
## Project Robyn

Add Unit Tests for `response_curve` pareto method

Fixes # (issue)

Fix this line: 
`metric_loc = slice(end_rw - n_periods + 1, end_rw)`
to
`metric_loc = slice(end_rw - n_periods, end_rw + 1)`

eg, date range is:

`[1,2,3,4,5]`

`rolling_window_start_which = 1`
`rolling_window_end_which = 5`

`slice(1,5)` would return `[1,2,3,4]` and not `[1,2,3,4,5]`

## Test Plan Results
image

image

### DIFF
--- a/python/src/robyn/modeling/pareto/response_curve.py
+++ b/python/src/robyn/modeling/pareto/response_curve.py
@@ -239,7 +239,7 @@ class ResponseCurveCalculator:
             date_range_updated = all_dates[metric_loc]
         elif date_range.startswith("last_"):
             n_periods = int(date_range.split("_")[1])
-            metric_loc = slice(end_rw - n_periods + 1, end_rw)
+            metric_loc = slice(end_rw - n_periods, end_rw)
             date_range_updated = all_dates[metric_loc]
         elif isinstance(date_range, list) and len(date_range) == 2:
             start_date, end_date = pd.to_datetime(date_range)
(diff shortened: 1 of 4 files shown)


python/src/robyn/modeling/pareto/response_curve.py
```diff
@@ -239,7 +239,7 @@ def _check_metric_dates(
             date_range_updated = all_dates[metric_loc]
         elif date_range.startswith("last_"):
             n_periods = int(date_range.split("_")[1])
-            metric_loc = slice(end_rw - n_periods + 1, end_rw)
+            metric_loc = slice(end_rw - n_periods, end_rw + 1)
```
alxlyj, line 242: will this break the pareto notebook?

alxlyj (approved):
updated the test plan with coverage numbers. I noticed that the test markdown content deviates a little from test_response_curve.py - is this expected behavior? Perhaps we can follow up a bit more on this.  

Overall, this looks good to me.

python/src/robyn/modeling/pareto/response_curve.py
Marco-Premier: Actually also my suggestion was wrong @alxlyj .

As the `slice(start,stop)` method in python, `start` is inclusive and `stop` is not as described [here.](https://docs.python.org/3/library/stdtypes.html#sequence-types) I'd keep the same pattern also for `rolling_window_start_which` and `rolling_window_end_which`, which implies this code:

`metric_loc = slice(end_rw - n_periods, end_rw)`

I also fix this line:
`date_range.startswith("last_"):` by adding a check first: `elif isinstance(date_range, str) and ` as if one passed the date_range as a list, the code thrown an exception

alxlyj (approved):
Thanks for making the changes.

PR #1104, opened 2024-11-01
One pager reporting implementation along with the plots.

sumalreddy17:
Added new OnePagerReporter class to create comprehensive one-page visualization reports 
combining multiple plot types from different visualizers.


dhavalpatel624624:
@sumalreddy17  can this be deleted? I think its old

dhavalpatel624624 (changes requested):
May be out of date... lets delete it

sumalreddy17:
outdated

PR #1106, opened 2024-11-04
added clustering tutorial and changed model entities

dhavalpatel624624:
## Project Robyn
[added clustering tutorial and changed model entities]
Added an enum for Hyperparameter_names
Updated data_mapper with new ClusteredResult Object mapping
Updated notebooks for pareto to use the new datamapper


python/src/robyn/modeling/entities/enums.py
```diff
@@ -27,3 +31,6 @@ class Models(Enum):
 
     RIDGE = auto()  # Ridge Regression
     # Add other model types as needed
+
+
+HYPERPARAMETER_NAMES = ("thetas", "shapes", "scales", "alphas", "gammas", "penalty")
```
sumane81, line 36: Can you define formal Enum for this? What you currently have is Constant more than Enum

dhavalpatel624624: I actually need is a constant not an enum (this is supposed to be used multiple files but many places has it implemented individually. Want it in a different file called constants?

sumane81 (approved):
Looks good to me.

PR #1108, opened 2024-11-05
Cluster_builder implementation

dhavalpatel624624:
Implementation logic for cluster_builder which clusters all the possible models and their results and selects the top ones by running a Kmeans on the clusters. 
Plotting is only included for wss (within groups sum of squares)
Plots for top_solutions_errors_plot, top_solutions_rois_plot, and clusters_confidence_interval_plot are PENDING

Testing Validation:
WSS Plot:
image

Confidence interval df and clusters mean values:
image

cluster_results, cluster data and the top solutions/models from the output:
image


python/src/robyn/modeling/clustering/cluster_builder.py
```diff
@@ -1,10 +1,38 @@
...
+
+import re
+import numpy as np
 import pandas as pd
-from robyn.modeling.clustering.clustering_config import ClusteringConfig
-from robyn.modeling.entities.clustering_results import ClusteredResult
+from robyn.modeling.clustering.clustering_config import (
+    ClusterBy,
+    ClusteringConfig,
+    DependentVarType,
```
alxlyj, line 14: shouldn't this be from robyn.data.entities.enums import DependentVarType?

python/src/robyn/tutorials/tutorial7_clustering.ipynb
alxlyj, line 1: could you address this?

A value is trying to be set on a copy of a slice from a DataFrame.
Try using .loc[row_indexer,col_indexer] = value instead

alxlyj (approved):
few nits

python/src/robyn/modeling/clustering/cluster_builder.py
dhavalpatel624624: good catch, I'll address it. I just took what it auto suggested.

python/src/robyn/tutorials/tutorial7_clustering.ipynb
dhavalpatel624624: Yep, will address it in the next PR :)

PR #1109, opened 2024-11-05
Test/unit tests pareto optimizer

Marco-Premier:
## Project Robyn

Add tests for pareto_optimizer

# Test Plan
testresult

test_coverage

### DIFF
--- a/python/spec/src/robyn/modeling/pareto/pareto_optimizer.md
+++ b/python/spec/src/robyn/modeling/pareto/pareto_optimizer.md
@@ -1,0 +1,184 @@
+# CLASS
+## ParetoResult
+* Holds the results of Pareto optimization for marketing mix models.
+* It is a data class that stores several attributes related to Pareto-optimal solutions.
+* Attributes:
+  - `pareto_solutions (List[str])`: A list of solution IDs that are Pareto-optimal.
+  - `pareto_fronts (int)`: The number of Pareto fronts considered in the optimization.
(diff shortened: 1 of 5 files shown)


dhavalpatel624624:
Can you add the pytest run results? Also are we going to be using the unittest library or pytest? We should standardize the libraries we use across all unit tests. Also, I think this needs a rebase on robynpy_release (you are out of date currently)

Also, are there any common best practices that should be followed here for generating mock data. I feel like there is a lot of redundancy in how the data is mocked for each function

Marco-Premier:
> Can you add the pytest run results? Also are we going to be using the unittest library or pytest? We should standardize the libraries we use across all unit tests. Also, I think this needs a rebase on robynpy_release (you are out of date currently)
> 
> Also, are there any common best practices that should be followed here for generating mock data. I feel like there is a lot of redundancy in how the data is mocked for each function

Goog points.
I'm happy to use pytest.
I refactor the code to use pytest and to group common mock data setup using pytest fixture.
Also I noticed that in tests where we aim to use pytest we still depend on unittest framework, eg. for `patch`, `MagicMock` etc...

We may consider using the `pytest-mock` instead.
I'd do that on a separate PR

dhavalpatel624624:
Still waiting for a rebase on this

dhavalpatel624624 (changes requested):
need to remove pareto_optimizer and response_curve file changes. A rebase should fix it :)

PR #1110, opened 2024-11-05
[test] add feature eng unit test and refactor unit test dir. 

alxlyj:


python/tests/modeling/test_feature_engineering.py
sumane81, line 1: Are we consistent on using testing framework? unittest vs pytest?

python/tests/modeling/test_feature_engineering.py
```diff
@@ -1,329 +1,412 @@
...
-            self.mmm_data.data = original_data
-            self.holidays_data.prophet_vars = original_prophet_vars
-
-if __name__ == '__main__':
-    unittest.main()
\ No newline at end of file
+            self.feature_engineering._set_holidays(dt_transform, self.holidays_data.dt_holidays, "invalid")
+
+
+if __name__ == "__main__":
```
sumane81, line 411: I prefer not to make individual test file runnable python class.

sumane81 (changes requested):

python/tests/modeling/test_feature_engineering.py
alxlyj: this uses pytest. the github workflows are all pytest - I think we could stick with pytest.

sumane81 (approved):

PR #1113, opened 2024-11-06
Robyn one pager: generate_diagnostic_plot implementation

sumalreddy17:
**Note**: Plotting is done with 90% accuracy expectation. Once all the 8 plots are generated and merged, a different PR will be raised if needed to take care of accuracy.

Add diagnostic plot for fitted vs residual analysis

Implements scatter plot of fitted values vs residuals
Adds LOWESS smoothing with confidence bands
Includes horizontal reference line
Matches R implementation styling
Uses consistent value formatting
Supports both standalone and subplot integration

Testing:
Screenshot 2024-11-05 at 11 27 21 PM

Graph plotted by R:
Screenshot 2024-11-06 at 2 54 06 PM

### DIFF
--- a/python/src/robyn/visualization/pareto_visualizer.py
+++ b/python/src/robyn/visualization/pareto_visualizer.py
@@ -1,4 +1,7 @@
+from typing import Optional
+from matplotlib import ticker
 import matplotlib.pyplot as plt
+import numpy as np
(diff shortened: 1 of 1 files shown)


sumane81 (approved):
Assuming accuracy testing and refactoring, if needed, will be done during integration testing.

PR #1114, opened 2024-11-06
generated base_visualizer and allocator_plotter using ai-codegen

Marco-Premier:
## Project Robyn

- generated base_visualizer.py using ai-codegen
- - generated allocator_plotter.py using ai-codegen

## Type of change
- Replaced existing `allocator_visualizer.py` (now renamed into `allocator_visualizer_original.py`) with a generated version from markdown.
- ai-codegen supports className == fileName, that's why I renamed the file `allocator_plotter.py`


python/src/robyn/visualization/allocator_plotter.py
```diff
@@ -0,0 +1,328 @@
...
+        x = np.arange(len(channels))
+
+        spend_pct = ((df["optimal_spend"] / df["current_spend"]) - 1) * 100
+        colors = [
+            self.positive_color if pct >= 0 else self.negative_color
+            for pct in spend_pct
+        ]
+        ax1.bar(x, spend_pct, color=colors, alpha=0.7)
+        ax1.set_xticks(x)
+        ax1.set_xticklabels(channels, rotation=45, ha="right"))
```
alxlyj, line 229: this has an extra parentheses

python/src/robyn/visualization/allocator_plotter.py
```diff
@@ -0,0 +1,328 @@
...
+            )
+
+        response_pct = ((df["optimal_response"] / df["current_response"]) - 1) * 100
+        colors = [
+            self.positive_color if pct >= 0 else self.negative_color
+            for pct in response_pct
+        ]
+        ax2.bar(x, response_pct, color=colors, alpha=0.7)
+        ax2.set_xticks(x)
+        ax2.set_xticklabels(channels, rotation=45, ha="right"))
```
alxlyj, line 249: this too.

python/src/robyn/visualization/allocator_plotter.py
```diff
@@ -0,0 +1,328 @@
...
+        self.result = result
+        plt.style.use("bmh")
+        plt.rcParams.update(
+            {
+                "figure.figsize": (12, 8),
+                "axes.grid": True,
+                "axes.spines.top": False,
+                "axes.spines.right": False,
+            }
+        )
```
alxlyj, line 21: looks like sizes are individually defined here within this component instead of base viz. I presume we will look at other common properties in next PR?

alxlyj (commented):
While AllocationPlotter extends BaseVisualizer, right now I don't see allocator leveraging most of the parent class functionality. 

**Inheritance usage:**

- It's redefining many properties already available in BaseVisualizer like colors, figure sizes
- It's even redoing style setup in __init__ that's already handled by parent class

**Style inconsistency:**

- Parent class has a structured color scheme but child overrides with its own simplified version
- Parent class has organized font sizes but child doesn't utilize them
- Child redefines rcParams that are already set in parent's _setup_plot_style()

**Methods** 

- we are not utilizing the parent's methods - such as save_plot() 

I will try to make some changes on my end.

alxlyj:
I made some changes here and also address the failing unit test checks as part of the github workflow:

- Proper inheritance and initialization:
`
def __init__(self, result: AllocationResult):
    super().__init__(style="bmh")  # Correctly calls parent init
`

alxlyj (approved):

sumane81 (approved):
Looks good.

PR #1115, opened 2024-11-06
[refactor] refactor modeling code

alxlyj:
### Summary

- Move modeling data entities to modeling/entities.
- Move feature engineering data entities to modeling/entities


sumane81 (approved):
Looks good to me.

PR #1123, opened 2024-11-08
Robyn one pager: Response curve plot implementation

sumalreddy17:
Add response curves visualization

Implement response curves showing spend vs response relationship
Add carryover effect shading for each channel
Display mean spend points with formatted labels
Support both standalone and subplot integration
Use abbreviated axis labels for large numbers
Add legend with channel colors
Include proper grid and axis styling

Testing:
Screenshot 2024-11-07 at 5 32 23 PM

Graph plotted by R code

Screenshot 2024-11-06 at 2 54 02 PM

### DIFF
--- a/python/src/robyn/visualization/response_visualizer.py
+++ b/python/src/robyn/visualization/response_visualizer.py
@@ -1,11 +1,14 @@
+from typing import Optional
 import matplotlib.pyplot as plt
+import numpy as np
+import pandas as pd
 from robyn.data.entities.mmmdata import MMMData
 from robyn.modeling.entities.pareto_result import ParetoResult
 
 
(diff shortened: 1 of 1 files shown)


sumane81:
Although we have plan to address accuracy later, testing plot and plot from R differs a lot. Is because data is different?

sumalreddy17:
Yes i believe the data used is different. Only way to verify is to use same data for both python and R code and plot graphs which is part of sanity check plan.

dhavalpatel624624:
If the testing is done on different dataset then this looks good to me. Approved.

dhavalpatel624624 (approved):

PR #1126, opened 2024-11-08
Cores update

sumane81:
Added CommonUtils with get_cores_available method and updated code to use it.


alxlyj (approved):
Looks good to me.

PR #1132, opened 2024-11-12
Exposure fitting, curve calibration & reach and frequency allocator

gufengzhou:
## Type of change
- feat: Exposure fitting is being reactivated. When paid_media_vars are different than paid_media_spend, paid_media_vars will be used for fitting. The extrapolation from exposure to spend uses the metric CPE (cost per exposure) = spend / exposure. Respectively, Michaelis Menten function for spend-exposure fitting is deprecated, because it doesn't provide significant improvement in fitting, while being very error-prone. The missing nonlinear relationship between spend and exposure will be address by curve calibration.
- feat: the curve calibrator. The new robyn_calibrate feature works standalone and without having built a model in Robyn.  It consumes cumulative reach and frequency and exports parameter ranges for MMM saturation calibration. This allows users to provide any saturation curves as proxies or source of truth to improve Robyn's saturation parameter identification and narrow down sample the ranges for alpha and gamma.
- prototype: reach and frequency allocator allows users to find optimum reach and frequency combination for media planning. 
- Refactor: aligned and simplified nonlinear transformation process across functions like run_transformation, robyn_response, robyn_pareto and robyn_allocator
- update of documentation & website and various bugs


igorskokan (approved):
Lgtm

igorskokan (approved):
Lgtm

PR #1135, opened 2024-11-13
clustering Plot fixes with better accuracy and Pareto minor change

dhavalpatel624624:
## Project Robyn

As titled:
Pareto was setting pareto_solutions to a different value
Clustering had some accuracy problems with graphs so addressed it.
Clustering methodologies are different from R so general sentiment matches the R code but the number of clusters getting selected when not provided and the selected clusters in graphs is slightly different. However, since the right channels seems be showing the similar results, I'm feeling comfortable with where we are with Clustering.

top_solutions_errors_plot:
image

top_solutions_rois_plot:
image

R equivalent
top_solutions_errors_plot:
image

top_solutions_rois_plot:
image


alxlyj (approved):
Looks good to me.

PR #1139, opened 2024-11-14
Integrated pareto clustering and visualizers

dhavalpatel624624:
## As titled, This PR integrates the pareto and clustering with visualizers one pager.

Added the utils function to map the pareto results after clustering 
Added a private method in pareto_visualizer to get the baseline variables
fixed minor edge cases which weren't working for bootstrap graph
changed solID to snake case sol_id

PENDING and Observations from testing:
-   I saw the actual vs predicted in R is generating some vertical lines which represent something, 
-   Adstock graph, bootstrap, and spend share vs effect share and has a simple sorting that needs to be done on the Y axis if you look (doesn’t have to match R but should be consistent for readability sake for users),
-   The x axis for all the response curves is really large compared to R which makes the graph look incorrect but I know its not
-  Fitted vs Residual graph looks incorrect in terms of the red line being drawn and the shaded area needs to tighter
-   The Response Decomponsition Waterfall graph should have the numbers closer to teh actual blue line, right now they are floating all over lol but the data and graph definitely looks correct


sumalreddy17 (approved):
The visualizer code changes LGTM.

sumane81 (approved):
Looks good. Thanks for noting down the observation and pending items.

PR #1141, opened 2024-11-15
[integration/accuracy]  enable clustering workflow and fix scalings for accuracy. 

alxlyj:
### Summary
- Inputs
  - Fix wrong instantiation on MMMData on all_media. 
  - Fix wrong instantiation on window_start and rolling_windows
- Modeling
  - Added scaling factor to match R's RSSD range
  - Better handling of coefficient unscaling
  - Proper zero effect detection with tolerance
  - Testing and comparisons with R exported data
  - Logger.debugs
  - Corrected R-style standard deviation calculation
  - Added proper scaling of data before lambda calculation
 
### Test Plans


python/src/robyn/modeling/ridge_model_builder_fail_pareto.py
sumane81, line 1: Can you remove backup/alternate files?

PR #1143, opened 2024-11-16
Update base_visualizer.py

sumane81:
Added abstract and static methods in base visualizer for common functionality


alxlyj (approved):
Followed up offline on more pending changes. Stamping to unblock.

PR #1145, opened 2024-11-16
Refactored robyn.py

sumane81:
Refactored robyn.py to have consistency in the way visualizer is used.

NOTE: this will break tutorial1. Subsequent PR will fix it.


alxlyj (approved):
This will fail tutorial1 but I'll put up a PR to fix them on tutorial1. Looks good.

PR #1146, opened 2024-11-18
Visualizer inheritance, basevisualizer and plot_all with working tutorial1

sumane81:
… BaseVisualizer

## Project Robyn

**BaseVisuazlier**
1. centralized display plot functionality
2. centralized export plot functionality - need to implement and test it
3. clean up

**All Visualizer**
1. Now extends BaseVisualizer and uses display from BaseVisualizer
2. Added plot_all method in all visuzliazers and update individual plot methods to maintain consistency.

**Robyn Client Interface (robyn.py)**
1. Enabled plots in all client interface apis

** Tutorial1 **
Updated and tested tutorial1 for all visualizers

**Not Yet Implemented/Tested**
1. ts_validation_plot is currently commented out to unblock. 
2. model convergence visualizer may need some updates
3. generate one pager in robyn client interface may need to make as instance method instead of static one.
4. tutorial1_src worked with local changes but didn't commit any changes in this PR. 
5. All other component level tutorials will need to be tested and may need updates for visualizer related changes


alxlyj (approved):
Looks like there are quite a few other changes that we need to address. Stamping first so we could move on to simplify other PR merges.

PR #1150, opened 2024-11-18
Robynpy bugfix graphs

dhavalpatel624624:
## Fixes robyn py in how it selects model before calling allocator
fixed the plot all function to display the cluster plot also
fixed a minor fix to pareto_optimizer where one list was getting set as a set instead of list

Tested with tutorial1 notebook

image


sumalreddy17 (approved):
LGTM

PR #1160, opened 2024-11-22
Robyn one pager end to end flow for tutorial 1

sumalreddy17:
Fix end-to-end one pager generation flow

* Remove redundant parameter passing
* Fix method call in plot_one_pager

- Update type hints and parameters
* Change List[str] to List[PlotType] for plots parameter
* Update ClusteringConfig to ClusteredResult
* Add proper return type hints

- Improve error handling
* Add better logging messages
* Add proper figure return handling

 Add single solution selection option

- Allow passing specific solution_id to plot_one_pager
* Maintain default top pareto behavior
* Update documentation with usage examples"

Testing: Verified one pager generation from Robyn class works end-to-end"

Screenshot 2024-11-22 at 10 16 22 AM

### DIFF
--- a/python/src/robyn/reporting/onepager_reporting.py
+++ b/python/src/robyn/reporting/onepager_reporting.py
@@ -404,7 +404,7 @@ class OnePager:
         Returns:
             List[plt.Figure]: List of generated figures, one per solution
         """
-         # Convert string plot types to PlotType if necessary
+        # Convert string plot types to PlotType if necessary
         if plots and isinstance(plots[0], str):
             try:
                 plots = [PlotType[plot.upper()] for plot in plots]
--- a/python/src/robyn/reporting/onepager_reporting.py
+++ b/python/src/robyn/reporting/onepager_reporting.py
@@ -219,27 +219,27 @@ class OnePager:
         """Add title and metrics text to the figure."""
         try:
             model_info = self._get_model_info(solution_id)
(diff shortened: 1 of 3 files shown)


python/src/robyn/robyn.py
```diff
@@ -398,29 +400,49 @@ def optimize_budget(
             logger.error("Budget optimization failed: %s", str(e))
             raise
 
-    @staticmethod
-    def generate_one_pager(
-        pareto_result: ParetoResult,
-        cluster_result: Optional[ClusteringConfig] = None,
-        mmm_data: Optional[MMMData] = None,
-        plots: Optional[List[str]] = None,
+    def plot_one_pager(self, solution_id: Optional[str] = None) -> None:
```
sumane81, line 403: This wrapper method on top of _generate_one_pager doesn't do anything. I would recommend just make generate_one_pager method public. Also, let's plot all plots from this method. No need to keep plotType input parameter here in this interface.

python/src/robyn/reporting/onepager_reporting.py
sumane81, line 1: See some formatter changes. Can you confirm if you are using Black Formatter in VSCode for formatting this?

sumalreddy17: yes, confirmed that it is formatted using black formatter.

sumane81 (approved):
Looks great.

PR #1169, opened 2024-12-02
Robynpy readme updates

dhavalpatel624624:


alxlyj:
@dhavalpatel624624  do you think we can add the Robyn R version where the Python code is based off of?

alxlyj (changes requested):
@dhavalpatel624624  do you think we can add the Robyn R version where the Python code is based off of?

PR #1175, opened 2024-12-05
[accuracy] modeling changes for calculate_decomp_spend_dist

alxlyj:
### Summary
- Added absolute value calculation for effect: effect = np.abs(coef * spend)
- Using absolute values for non-zero effect calculation: non_zero_effect = np.abs(X[col][non_zero_mask] * coef)
- Effect shares are now calculated using these absolute effects

image


dhavalpatel624624 (approved):
Looks good to me!

PR #1180, opened 2024-12-06
Code cleanup and proper logging for pareto and clustering

dhavalpatel624624:
As titled... tested using tutorial.ipynb and still failing at the negative values as expected from robynpy_release


alxlyj (approved):
looks good to me - nit: we probably want to change the call or the name to RobynLogger, but I think we should table this to the next batch of changes.

PR #1181, opened 2024-12-06
[cleanup] cleanup loggers for console/notebook

alxlyj:
### Summary
- As titled, we are cleaning up loggers and visual display outputs.


python/src/robyn/modeling/feature_engineering.py
```diff
@@ -19,6 +19,15 @@
 from robyn.data.entities.hyperparameters import Hyperparameters, ChannelHyperparameters
 from robyn.data.entities.mmmdata import MMMData
 
+# Set logging level for this module
+logging.getLogger("cmdstanpy").setLevel(logging.CRITICAL)  # Only show critical errors
```
sumane81, line 23: What does this and next line do? logger for this class is instantiated later, which will not affect with these two lines

python/src/robyn/visualization/base_visualizer.py
```diff
@@ -14,6 +14,7 @@
 
 # Configure logger
 logger = logging.getLogger(__name__)
+logger.setLevel(logging.WARNING)
```
sumane81, line 17: Use should be able to control this. Setting log level should be at one time config level only.

python/src/robyn/visualization/feature_visualization.py
```diff
@@ -10,8 +10,15 @@
 from robyn.data.entities.hyperparameters import Hyperparameters
 from robyn.modeling.feature_engineering import FeaturizedMMMData
 from robyn.visualization.base_visualizer import BaseVisualizer
+import warnings
+
+# Set logging level for this module
+logging.getLogger("cmdstanpy").setLevel(logging.CRITICAL)  # Only show critical errors
+logging.getLogger("prophet").setLevel(logging.CRITICAL)  # Only show critical errors
```
sumane81, line 17: Not sure what these two lines will do

sumane81 (approved):
Looks good.

PR #1192, opened 2024-12-20
fix: adstock plot warning on changing geom_text() to annotate()

laresbernardo:
Fixing the following warning everytime a one-pager is created:

```
Warning in geom_text(aes(x = max(.data$x), y = 0.5, vjust = -0.5, hjust = 1, :
All aesthetics have length 1, but the data has 250 rows.
ℹ Please consider using annotate() or provide this layer with data containing a single row.
```

The warning message is due to the fact that geom_text is trying to apply the same label ("Halflife") across multiple rows of the data, which causes a mismatch in the number of aesthetics.


gufengzhou (approved):
LGTM

PR #1193, opened 2024-12-23
fix: min max bug when calibration returns NaN + underlying linetype data on actual vs predicted plot

laresbernardo:


gufengzhou (approved):
LGTM

PR #1194, opened 2024-12-23
fix: prev versions' response curves and mean spends per channel on one-pagers

laresbernardo:


laresbernardo:
Happy New Year @gufengzhou :) 
This one is ready to land. Fixes the previous version models issue while rendering one-pagers.

gufengzhou:
> Happy New Year @gufengzhou :) This one is ready to land. Fixes the previous version models issue while rendering one-pagers.

Does it work with paid_media_vars too? If yes I'd prefer that

gufengzhou (approved):
LGTM

PR #1199, opened 2025-01-09
fix: modeling window coverage considers all actual dates data + legacy InputCollect for BA bug

laresbernardo:
Improve modeling window coverage accuracy by considering all real dates in the data. With this change:

1. When window_end is not defined, the default is now the latest available day before the next logical date in the data. 
2. Allows for days larger than the max but smaller than the next logical date without a warning.

The results are exactly the same, it's a matter of "correct" coverage. For monthly data, if modeling window covers from "2024-01-01" to "2024-12-31", the data for last month is "2024-12-01" but it actually covers until "2024-12-31".

This logic is triggered only when window_end is not provided.

---

Also, fixed issue with legacy InputCollect and budget allocator using paid_media_selected (similar to #1194)

### DIFF
--- a/R/R/allocator.R
+++ b/R/R/allocator.R
@@ -133,7 +133,8 @@ robyn_allocator <- function(robyn_object = NULL,
 
   ## set local variables, sort & prompt
   # paid_media_spends <- InputCollect$paid_media_spends
-  paid_media_selected <- InputCollect$paid_media_selected
+  paid_media_selected <- if ("paid_media_selected" %in% names(InputCollect))
+    InputCollect$paid_media_selected else InputCollect$paid_media_spends
   dep_var_type <- InputCollect$dep_var_type
   if (is.null(select_model) && length(OutputCollect$allSolutions == 1)) {
     select_model <- OutputCollect$allSolutions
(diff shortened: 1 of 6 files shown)


R/R/allocator.R
```diff
@@ -133,7 +133,8 @@ robyn_allocator <- function(robyn_object = NULL,
 
   ## set local variables, sort & prompt
   # paid_media_spends <- InputCollect$paid_media_spends
-  paid_media_selected <- InputCollect$paid_media_selected
+  paid_media_selected <- if ("paid_media_selected" %in% names(InputCollect))
+    InputCollect$paid_media_selected else InputCollect$paid_media_spends
```
gufengzhou, line 137: For this change, would you please try simply replacing `InputCollect$paid_media_selected` with `InputCollect$paid_media_vars`. If that works, I'd prefer that. Otherwise looks good.

gufengzhou (changes requested):
See my comment reg. the allocator.R change.

R/R/allocator.R
laresbernardo: Ok, so for the one-pagers we can use `InputCollect$paid_media_vars` directly, no need to check if it exists because it's legacy. What is new is `paid_media_selected`. Changing to `InputCollect$paid_media_vars` then and getting rid of this check.

laresbernardo: @gufengzhou checked and working with legacy results as well.

gufengzhou (approved):
lgtm

PR #1201, opened 2025-01-09
Remove duplicate nevergrad dependency

pniskala:
## Project Robyn

Looking at the `pyproject.toml` and `requirements.txt` files, `nevergrad` is listed twice as a dependency. This PR removes the duplicate dependency from both files. This should only be a cosmetic change.

## Type of change
- remove duplicate nevergrad from requirements.txt for robynpy
- remove duplicate nevergrad from pyproject.toml for robynpy


sumane81 (approved):
LGTM

PR #1212, opened 2025-02-04
feat: robyn_response()'s plot labels improvement

laresbernardo:
- Adding total input and calculation results when date range or metric value are provided into `robyn_response()`.
- For further clarification, added time granularity in x axis and dep variable in y axis.

PENDING BEFORE MERGING:
- [ ] It seems there's an issue on counting the periods in some cases. When testing `c("2022-06-01", "2023-05-01")` for example, I get 2 periods, which is incorrect.
- [ ] Strange results are showing up sometimes. When `metric_value = 0`, I see the blue dot is larger than the grey dot, which is historical - which doesn't make sense. Maybe fixing the prior point fixes this as well.
- [ ] What's the difference between `sum(resp$input_total)` and `metric_value`? @gufengzhou

### DIFF
--- a/R/DESCRIPTION
+++ b/R/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: Robyn
 Type: Package
 Title: Semi-Automated Marketing Mix Modeling (MMM) from Meta Marketing Science 
-Version: 3.12.0.9006
+Version: 3.12.0.9007
 Authors@R: c(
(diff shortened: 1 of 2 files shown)


laresbernardo:
Example of `metric_value = 0`  result.
image

Example of `metric_value = NULL` result.
image

PR #1226, opened 2025-02-25
[fix] robynpy modeling component

alxlyj:
### Type of change
- Fixes for modeling convergence.
- Included logging at modeling steps to compare with R. (TODO: cleanup logging in separate PR)
- Updated README to include setup for R installation for glmnet for rpy2.

### Testing

image


python/src/robyn/modeling/ridge/models/ridge_utils.py
sumane81, line 1: To be consistent, I would prefer we make this object oriented by defining a class. It's okay if you want to do this in follow up PR.

sumane81 (approved):
Make ridge utils a class and any additional clean up in following PR. Approving this to unblock.

leonl0000:
I believe the line adding "newsletter" in `robyn/modeling/ridge/ridge_data_builder.py` causes it to crash anytime there's no newsletter in the data @alxlyj

PR #1238, opened 2025-03-19
[fix] onepager and budget allocator fixes

alxlyj:
## Project Robyn
- Fixes for onepager and budget allocator. 

## Testing
- Ran tutorial1.ipynb, tutorial1_src.ipynb. 
image

image


python/pyproject.toml
```diff
@@ -34,6 +34,8 @@ dependencies = [
     "nlopt",
     "ipykernel",
     "rpy2==3.5.16",
+    "kaleido==0.2.1",
+    "nlopt==2.9.1",
```
sumane81, line 38: Are we now using nlopt?

python/requirements.txt
```diff
@@ -13,4 +13,6 @@ plotly==5.24.1
 pytest==8.3.3
 nlopt==2.8.0
 ipykernel==6.29.5
-rpy2==3.5.16
\ No newline at end of file
+rpy2==3.5.16
+kaleido==0.2.1
```
sumane81, line 17: What does this do?

python/src/robyn/allocator/checks.py
sumane81, line 1: All of these are validations. Can you follow standard validation that we are doing? This breaks the entire architecture. I don't see class definition, validation result etc.

python/src/robyn/data/validation/checks.py
sumane81, line 1: Again, this completely breaks the pattern.

sumane81 (changes requested):
I see multiple evidences that this change completely breaks modular architecture. It was our efforts to not end up in mess the way R code is. If functionality is working fine, please refactor code to follow the standard for organizing code.

python/pyproject.toml
alxlyj: yes

python/requirements.txt
alxlyj: helps output plots for different figures

python/src/robyn/allocator/checks.py
alxlyj: I will address this in a further change - not all check methods has been included in our first rewrite for checks.R so it's not difficult to add those in.

alxlyj:
yes, I am aware that the patterns here are different from what we already have in robynpy since accuracy and comparisons were my focus here. I was going to further address this in a separate PR, but I'll make a few more changes here to refactor the code.

sumane81 (approved):
Looks good. thanks for addressing the concerns!

PR #1284, opened 2025-06-27
fix: refresh window end issue fixes #1270

### LINKED ISSUE #1270: Incorrect `window_end` when using `robyn_read(json_file)`
Hi Robyn team,

I found a potential issue when reloading a model using `robyn_read(json_file)`.

In my case:
- The raw data ends at **2025-04-14**, confirmed via `json_file$Extras$raw_data$date`
- But `json_file$InputCollect$window_end` is incorrectly set to **2025-04-20**

max(json_file$Extras$raw_data$date)
 [1] "2025-04-14"   # <-- correct
json_file$InputCollect$window_end
 [1] "2025-04-20"   # <-- incorrect

This causes a refresh issue. Even after adding the next week of data, I get the following error:
-> Not enough data for this refresh. Input data from date 2025-04-27 or later required

It looks like window_end is not correctly reconstructed, and that prevents proper model refresh even when enough data is available

Questions: 
1. Is this a known issue with robyn_read() incorrectly setting window_end?
2. Is the refresh feature under active development? It feels somewhat unstable currently.

Robyn Version 3.12.0.9006

Thanks as always for your work and support on Robyn!
@gufengzhou

laresbernardo:
This fix has been [checked and confirmed](https://github.com/facebookexperimental/Robyn/issues/1270#issuecomment-3012069076) by user.

Last PR before submitting latest dev version to CRAN.


gufengzhou (approved):
LGTM

